PR #3, opened 2020-04-27
Add a layer to record time between calls to vkQueuePresentKHR

brianwatling:
This can be used to measure raw frame time

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 This project contains two Vulkan layers:
 1. Compile-time layer for measuring pipeline compilation times.
 2. Runtime layer for measuring pipeline execution times.
+3. Frame time layer for measuring time between calls to vkQueuePresentKHR, in nanoseconds.
 
 The results are saved as CSV files.
 
--- a/layer/layer_data.cc
+++ b/layer/layer_data.cc
@@ -14,6 +14,7 @@
 
 #include "layer_data.h"
 
+#include <inttypes.h>
 #include <cstdint>
 
 #include "absl/strings/str_cat.h"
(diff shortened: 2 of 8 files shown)


kuhar (changes requested):
Could you document what the output format is? At least in the commit description, but might be good to have that in markdown too.

Looks good to me otherwise.

README.md
```diff
@@ -3,6 +3,7 @@
 This project contains two Vulkan layers:
 1. Compile-time layer for measuring pipeline compilation times.
 2. Runtime layer for measuring pipeline execution times.
+3. Frame time layr for measuring time between calls to vkQueuePresentKHR, in nanoseconds.
```
kuhar: nit: layer

layer/layer_data.cc
```diff
@@ -14,6 +14,8 @@
 
 #include "layer_data.h"
 
+#include <inttypes.h>
```
kuhar, line 17: nit: Could you move this below the next include to keep them sorted?

layer/layer_data.h
```diff
@@ -45,7 +45,7 @@ class LayerData {
       absl::flat_hash_map<VkInstance, VkLayerInstanceDispatchTable>;
   using DeviceDispatchMap = absl::flat_hash_map<VkDevice, VkLayerDispatchTable>;
 
-  explicit LayerData(char* log_filename);
+  explicit LayerData(char* log_filename, const char* header);
```
kuhar: nit: do we need `explicit` if this constructor takes more than one argument?

kuhar (approved):

layer/layer_data.cc
brianwatling: clang-format reordered it this way (I guess because it's a c header while cstdint is a c++ header). I've reran clang-format with a different version and the blank line is now gone

PR #5, opened 2020-05-13
Add a rough script to plot frame times

brianwatling:


scripts/plot_frame_times.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
```
kuhar: Use `#!/usr/bin/env bash`

scripts/plot_frame_times.sh
```diff
@@ -0,0 +1,33 @@
...
+#
+# This script can be used to plot frame time layer output with gnuplot. A PNG is
+# generated containing a CDF of frame times in milliseconds.
+#
+# Usage:
+#   plot_fram_times.sh frame_time_a.txt frame_time_b.txt
```
kuhar: Can you mention that what the default output filename is?

scripts/plot_frame_times.sh
```diff
@@ -0,0 +1,33 @@
...
+
```
kuhar, line 9: Consider adding `set -e`

kuhar (approved):
Very useful. LGTM modulo some nits.

scripts/plot_frame_times.sh
brianwatling: Done

scripts/plot_frame_times.sh
brianwatling: Done

scripts/plot_frame_times.sh
brianwatling: Done

PR #8, opened 2020-05-29
[Frame Time] Detect when benchmark exits

kuhar:
This will make sure that the finish indicator file is created when the
benchmarked Vulkan application exists on its own.

Make sure that layers' logs are flushed after each print to avoid
creating corrupted logs when benchmark dies.

### DIFF
--- a/layer/compile_time_layer.cc
+++ b/layer/compile_time_layer.cc
@@ -40,10 +40,10 @@ constexpr char kLayerDescription[] =
 constexpr char kLogFilenameEnvVar[] = "VK_COMPILE_TIME_LOG";
 
 performancelayers::LayerData* GetLayerData() {
-  static performancelayers::LayerData* layer_data =
-      new performancelayers::LayerData(getenv(kLogFilenameEnvVar),
-                                       "Pipeline,Compile Time (ns)");
-  return layer_data;
+  // Don't use new -- make the destructor run when the layer gets unloaded.
+  static performancelayers::LayerData layer_data = performancelayers::LayerData(
+      getenv(kLogFilenameEnvVar), "Pipeline,Compile Time (ns)");
+  return &layer_data;
 }
 
 //////////////////////////////////////////////////////////////////////////////
--- a/layer/frame_time_layer.cc
+++ b/layer/frame_time_layer.cc
@@ -46,6 +46,8 @@ class FrameTimeLayerData : public performancelayers::LayerData {
       : LayerData(log_filename, "Frame Time (ns)"),
         exit_frame_num_or_invalid_(exit_frame_num_or_invalid) {}
 
+  ~FrameTimeLayerData() override;
+
   // Records the device that owns |queue|.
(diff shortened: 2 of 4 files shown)


layer/frame_time_layer.cc
```diff
@@ -103,33 +136,18 @@ VKAPI_ATTR VkResult FrameTimeLayer_QueuePresentKHR(
...
         "\n",
         frames_elapsed);
     fflush(stderr);
+
     // _Exit will bring down the parent Vulkan application without running any
     // cleanup. Resources will be reclaimed by the operating system.
-    // Delete to force the destructor to save the layer log files.
-    delete layer_data;
+    // Call the destructor manually to save the layer log files.
+    layer_data->~FrameTimeLayerData();
```
brianwatling: I'd prefer to see this call layer_data->CreateFinishIndicatorFile("FRAME_TIME_LAYER_TERMINATED") rather than manually calling the destructor. Then the destructor can also just be  CreateFinishIndicatorFile("APPLICATION_EXIT);

brianwatling (approved):

layer/frame_time_layer.cc
kuhar: Calling the destructor ensures that the log file is closed. But since every print is now followed by a flush, I guess this should be fine?

brianwatling: Correct, it's fine to _exit without closing the file because of the fflush() calls. Alternatively, consider making the static local variables unique_ptr's that you can reset here (that way you can manually trigger deletion here and when the layer is unloaded). Technically calling the destructor as in your original implementation is fine, but it's just surprising to see a direct destructor call for something that isn't created with placement new

layer/compile_time_layer.cc
```diff
@@ -40,10 +40,10 @@ constexpr char kLayerDescription[] =
 constexpr char kLogFilenameEnvVar[] = "VK_COMPILE_TIME_LOG";
 
 performancelayers::LayerData* GetLayerData() {
-  static performancelayers::LayerData* layer_data =
-      new performancelayers::LayerData(getenv(kLogFilenameEnvVar),
-                                       "Pipeline,Compile Time (ns)");
-  return layer_data;
+  // Don't use new -- make the destructor run when the layer gets unloaded.
+  static performancelayers::LayerData layer_data = performancelayers::LayerData(
+      getenv(kLogFilenameEnvVar), "Pipeline,Compile Time (ns)");
```
dnovillo, line 45: why not used some C++ managed pointer and use RAII?

kuhar: What difference would it make here? I think the point of using new in the past was to not register a global destructor as suggested by the c++ style guide, but in this case we do want to run the destructor when the layer gets unloaded.

kuhar: IE the lifetime on the meta-level is managed by the Vulkan loader.

dnovillo: ah, nm.  In that case, this is fine.  Thanks.

PR #13, opened 2020-06-17
Record fragment and compute shader invocations

brianwatling:
This change adds a query to record fragment and compute shader
invocations. Immediate calls to vkCmdResetQueryPool are removed since
vkCmdResetQueryPool "... must only be called outside of a render pass instance"


layer/layer_data.cc
```diff
@@ -90,6 +90,14 @@ void LayerData::Log(const std::vector<uint64_t>& pipeline,
   fflush(out_);
 }
 
+void LayerData::Log(const std::vector<uint64_t>& pipeline,
+                    const char* str) const {
```
kuhar: What do you think about just passing a std::string instead and using something less brittle than `snprintf` at the call sites, e.g., std::stringstream or absl::StrCat?

layer/runtime_layer.cc
```diff
@@ -169,41 +169,78 @@ VKAPI_ATTR void RuntimeLayer_CmdDispatch(VkCommandBuffer command_buffer,
...
   auto next_proc = layer_data->GetNextDeviceProcAddr(
       device, &VkLayerDispatchTable::CmdDispatch);
 
-  // Get the timestamp before the dispatch starts.
-  (write_timestamp_function)(command_buffer,
-                             VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT,
-                             timestamp_query_pool, 0);
+  VkQueryPool timestamp_query_pool = VK_NULL_HANDLE;
+  VkQueryPool stat_query_pool = VK_NULL_HANDLE;
+  if (!layer_data->GetNewQueryInfo(command_buffer, &timestamp_query_pool,
```
kuhar, line 178: when can we get a null here?

layer/runtime_layer_data.cc
```diff
@@ -105,14 +121,30 @@ void RuntimeLayerData::LogAndRemoveQueryPools() {
...
+
     if (!discard_result) {
       // FIXME: We should adjust the elapsed units to account for the current
       // GPU frequency. Calling vkGetPhysicalDeviceProperties here causes the
       // driver to crash, however.
-      Log(GetPipelineHash(info->pipeline), timestamp1 - timestamp0);
+      char buffer[256];
+      snprintf(buffer, sizeof(buffer), "%" PRIu64 ",%" PRIu64 ",%" PRIu64,
+               timestamp1 - timestamp0, invocations[0], invocations[1]);
+      buffer[sizeof(buffer) - 1] = 0;
```
kuhar: Maybe zero-initialize the whole buffer, or just stringstream or some formatters from absail?

layer/runtime_layer_data.cc
```diff
@@ -105,14 +121,30 @@ void RuntimeLayerData::LogAndRemoveQueryPools() {
...
+      Log(GetPipelineHash(info->pipeline), buffer);
```
kuhar: Doesn't this require us to also update the CSV header?

kuhar (commented):
I think we should start including sample logs in order to be quickly able to just quickly tell what the layers actually produce and in what format.

layer/runtime_layer.cc
kuhar: IE, can you add a comment saying when this can happen and why it's fine to just bail out

layer/layer_data.cc
brianwatling: Done

layer/runtime_layer_data.cc
brianwatling: switched to StrCat

layer/runtime_layer_data.cc
brianwatling: Done

brianwatling:
> I think we should start including sample logs in order to be quickly able to just quickly tell what the layers actually produce and in what format.

I've added some samples and a script to help generate them using vkcube

layer/runtime_layer.cc
brianwatling: done

kuhar (approved):

PR #16, opened 2020-09-11
Add LogScanner and use it for benchmark start detection

kuhar:
This adds a new support utility: `LogScanner`. The basic idea is
that the moment when games finish loading and start playing can
be detected by looking at log files, produced either by the game
itself, or some external tool (e.g., layer, strace, etc.).

`LogScanner` allows for watching log files and consuming new content
as it's continously written.

This PR adds a very basic benchmark start detection to the Frame Time
Layer. The Frame Time Layer should probaly consume a config file,
given the growing number of env vars, but this is out of scope of this
PR.

Also add unit testing support through googletest, and link with the
correct filesystem library (note: only libstdc++ and libc++ are supported
at the moment).

Test: run `ninja check`.


layer/frame_time_layer.cc
```diff
@@ -117,6 +146,23 @@ void CreateFinishIndicatorFile(const char* finishCause) {
...
 }
 
+bool FrameTimeLayerData::HasBenchmarkStarted() {
+  if (benchmark_state_idx_ > 0 || benchmark_start_pattern_.empty()) return true;
+
+  if (!benchmark_log_scanner_.has_value() || !benchmark_log_scanner_->is_open())
+    return true;
+
+  if (benchmark_log_scanner_->consume_new_lines()) {
+    benchmark_state_idx_ = 1;
```
brianwatling, line 154: a bool seems more appropriate if benchmark_state_idx_ can only be 0 or 1 (or was this supposed to receive the value of get_first_occurrence_line_num()?)

layer/log_scanner.h
```diff
@@ -0,0 +1,73 @@
...
+class LogScanner {
+ public:
+  // Opens |filename| and returns a LogScanner object on success,
+  // or std::nullopt on failure.
+  static std::optional<LogScanner> from_filename(const std::string& filename);
+
+  // Reads the underlying log file until the end. Checks if any of the
+  // registered patterns appear within the lines read.
+  // Returns true if any register pattern was found for the first time.
+  bool consume_new_lines();
```
brianwatling: nit: other code in this repo follows the Google style guide where member functions are CamelCase, better to be consistent I think

layer/log_scanner.h
```diff
@@ -0,0 +1,73 @@
...
+
+  // Registers |pattern| as a watched pattern. No-op if pattern has
+  // already been registered.
+  // Patterns are matched using exact (full) maching only.
+  void register_watched_pattern(const std::string& pattern) {
+    pattern_to_line_num_.try_emplace(pattern, 0);
+  }
+
+  // Stops watching |pattern| and forgets the first match, if present.
+  void unregister_watched_pattern(const std::string& pattern);
```
brianwatling: this isn't used in the layer or tests - please remove and add it in a separate patch when needed

layer/log_scanner.h
```diff
@@ -0,0 +1,73 @@
...
+  void close() { file_.close(); }
+
+  // Returns true iff the underlying file is open.
+  bool is_open() const { return file_.is_open(); }
+
+ private:
+  LogScanner(std::ifstream file) : file_(std::move(file)) {}
+  std::ifstream file_;
+  uint64_t current_line_num_ = 0;
+  absl::flat_hash_map<std::string, uint64_t> pattern_to_line_num_;
```
brianwatling, line 59: I don't think the layer needs to search for more than one pattern and does't need the line number (it only needs to know the pattern was found). Could the LogScanner API be simplified to something like the following?:
optional<LogScanner> scanner = LogScanner(file, pattern);
...
if (scanner.PatternFound()) {
  found_ = true;
  scanner.reset();
}

layer/frame_time_layer.cc
```diff
@@ -117,6 +146,23 @@ void CreateFinishIndicatorFile(const char* finishCause) {
...
+    benchmark_log_scanner_->get_first_occurrence_line_num(
+        benchmark_start_pattern_);
+    benchmark_log_scanner_->close();
```
brianwatling: IMHO exposing close() and is_open() pollutes the LogScanner API - the layer could just clear the optional<> when it's done with the scanner (see my other comment about simplifying the LogScanner API)

brianwatling (changes requested):
Looks ok, just have some comments to simplify the LogScanner a bit

layer/frame_time_layer.cc
kuhar: I intent to extend this to detecting multiple game states, e.g, initialization, in-menu, in-level, etc. While a `bool` makes sense now, I didn't want to have to refactor this in near future.

layer/log_scanner.h
kuhar: I want to extend it to search for multiple patterns in a future PR. I found line numbers to be useful for debugging: they allow for jumping to the actual log file and looking for the pattern in the surrounding context.

layer/frame_time_layer.cc
kuhar: Good suggestion, I like this idea

layer/log_scanner.h
kuhar: Done

layer/log_scanner.h
kuhar: Done

brianwatling (approved):

PR #18, opened 2020-09-22
Pass VkInstance to CreateDevice. Fixup intercepted function attributes.

kuhar:
CreateDevice needs to know the exact instance when there is more than
one layer active:
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/html/chap4.html#initialization-functionpointers.

Override EnumeratePhysicalDevices to keep a map from physical devices to
instances.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -63,6 +63,7 @@ target_include_directories(VkLayer_stadia_pipeline_compile_time PRIVATE
 )
 target_link_libraries(VkLayer_stadia_pipeline_compile_time PRIVATE
     absl::flat_hash_map
+    absl::flat_hash_set
     absl::strings
     absl::str_format
     absl::synchronization
@@ -85,6 +86,7 @@ target_include_directories(VkLayer_stadia_pipeline_runtime PRIVATE
 )
 target_link_libraries(VkLayer_stadia_pipeline_runtime PRIVATE
     absl::flat_hash_map
+    absl::flat_hash_set
     absl::strings
     absl::str_format
     absl::synchronization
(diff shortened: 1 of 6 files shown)


brianwatling (approved):

paulthomson:
Actually, it turns out there is a way to do this without tracking `VkPhysicalDevice` (in case you are interested). From my notes:

```
The following are dispatchable handles:

- VkInstance
- VkPhysicalDevice (same dispatch table as its VkInstance)

- VkDevice
- VkQueue (same dispatch table as its VkDevice)
- VkCommandBuffer (same dispatch table as its VkDevice)

Using the dispatch table pointer of such handles as keys means we don't need
to track physical devices, queues, or command buffers, unless we want to
store information about each of these objects.
```

In other words, instead of using `VkInstance` as a key (and the same idea applies for `VkDevice`), use the "dispatch table pointer" of the `VkInstance`, which you get using [this function](https://github.com/google/gf-layers/blob/98a9747b711838f49a296dada34ba9aa64346307/src/gf_layers_layer_util/src/util.cc#L21) (please ignore the snake_case naming). And when you have a `VkPhysicalDevice`, use [pretty much the same function](https://github.com/google/gf-layers/blob/98a9747b711838f49a296dada34ba9aa64346307/src/gf_layers_layer_util/src/util.cc#L25) to get the VkInstance dispatch table pointer, which you can then use to look-up the data you have for the `VkInstance` (which should include the `VkInstance` value itself). 

Another example of where this is somewhat nice is in a frame counter layer. We don't need to track `VkQueue`; [we can go straight from the VkQueue to the DeviceData](https://github.com/google/gf-layers/blob/98a9747b711838f49a296dada34ba9aa64346307/src/VkLayer_GF_frame_counter/src/frame_counter_layer.cc#L137).

On the other hand, tracking `VkPhysicalDevice`s, `VkQueue`s, etc. is also fine, especially if you need to track them anyway.

This advice is documented here:
https://github.com/KhronosGroup/Vulkan-Loader/blob/master/loader/LoaderAndLayerInterface.md#hash-maps

...although I feel it isn't explained very clearly.

kuhar:
Very cool, thanks @paulthomson. Always happy to have less work to do, we should switch to this method.

PR #21, opened 2020-09-28
Add an option to log events from all layers in a single file

kuhar:
This allows for matching events from all layers based on the time
when they happened. For example, hitches caused by pipeline creations
can be correclated with increased frame times.

Update the readme and sample outputs accordingly.


kuhar:
Fixed some typos

layer/layer_data.cc
```diff
@@ -72,15 +103,24 @@ LayerData::LayerData(char* log_filename, const char* header) {
...
-  fflush(out_);
+
+  WriteLnAndFlush(out_, header);
+
+  if (const char* event_log_file = getenv(kEventLogFileEnvVar)) {
+    absl::MutexLock lock(&event_log_lock_);
+    // We should probably use API that allows us to lock the event log file so
+    // that every line is written atomically and flushed. Printing and fflushing
+    // seems to be working fine at the moment, but we can revisit this in the
+    // future.
```
brianwatling: I'd remove or reword this comment to simply say all lines should be written in a single fprintf call (the FILE functions are atomic within a process, as in two fprintf's or other calls to the same FILE* will internally take a lock).

You've refactored all the fprintf() calls to use WriteLnAndFlush which ensures all writes are done in single fprintf calls - because of this log_lock_ and event_log_lock_ can be removed entirely (they could before as well AFAICT, but now it's clearer that all writes are full lines)

layer/layer_data.cc
```diff
@@ -72,15 +103,24 @@ LayerData::LayerData(char* log_filename, const char* header) {
...
   } else {
     out_ = stderr;
   }
-  fprintf(out_, "%s\n", header);
-  fflush(out_);
+
+  WriteLnAndFlush(out_, header);
+
+  if (const char* event_log_file = getenv(kEventLogFileEnvVar)) {
+    absl::MutexLock lock(&event_log_lock_);
```
brianwatling: no need to take this lock in the constructor since no other threads will be able to access this object until after the constructor completes

brianwatling (changes requested):
overall looks good

layer/layer_data.cc
kuhar: Done. Nice observation.

brianwatling (approved):

PR #22, opened 2020-09-30
Add a script for plotting event log files

kuhar:
This is useful for comparing different benchmark configurations
side by side and investigating causes of performance issues.
For example, it's easy to correlate loading times and hitches with pipeline
creations.


kuhar:
Sample output (also included in the PR):
image

README.md
```diff
@@ -2,14 +2,23 @@
...
 The results are saved as CSV files. Setting the `VK_PERFORMANCE_LAYERS_EVENT_LOG_FILE` environment variable makes all layers append their events (with timestamps) to a single file.
 
 The layers are considered experimental.
 We welcome contributions and suggestions for improvements; see [docs/contributing.md](docs/contributing.md).
 
+## Analysis Scripts
+
+The project comes with a set of simple log analysis scripts:
+1. [analyze_frametimes.py](scripts/analyze_analyze_frametimes.py) -- processes frame time layer logs. Prints summarized results, outputs frames per second (FPS) as CSV files, and plots frame time distributions.
+2. [plot_timeline.py](scripts/plot_timeline.py) -- processes even log files. Plots frames per second (FPS) and pipeline creation times. Sample output:
```
brianwatling: s/even/event/

scripts/plot_timeline.py
```diff
@@ -0,0 +1,185 @@
...
+import os.path
+import sys
+
+nanos_per_second = 10**9
+nanos_per_milli = 10**6
+
+def get_frames_per_second(frame_present_events):
+    first_second = frame_present_events[0][0] // nanos_per_second
+    last_second = int(np.ceil(frame_present_events[-1][0] / nanos_per_second))
+    duration_seconds = last_second - first_second + 1
```
brianwatling, line 50: I suggest leaving a comment saying this function buckets frames by second and what the return values are (at first glance I wondered why whole numbers are used, but it makes sense when bucketing by second). Would be good to do the same for some of the other functions as well

brianwatling (approved):

scripts/plot_timeline.py
kuhar: Thanks for the suggestion, done. The part of the code that calculates FPS is almost the same as in `analyze_frametimes.py`, so it might be worth factoring this out into a common library file in the future, should the code continue to grow.

PR #31, opened 2021-02-24
Fix open source logging. Move logging implementation to a cc file.

kuhar:
Fix open source logging. Move logging implementation to a cc file.

Introduce a simple helper class for logging. This makes sure that
the concrete logging library is hidden from layer files that can't
depend on implementation details.

Include source file names and line numbers in open source logs.

This depends on https://github.com/googlestadia/performance-layers/pull/30 and needs to be rebased before submitting.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,7 @@ message(STATUS "VulkanLoaderGenerated ${VulkanLoaderGenerated_INCLUDE_DIR}")
 add_library(performance_layers_support_lib INTERFACE)
 target_sources(performance_layers_support_lib INTERFACE
     ${CMAKE_CURRENT_SOURCE_DIR}/layer/layer_data.cc
+    ${CMAKE_CURRENT_SOURCE_DIR}/layer/logging.cc
     ${CMAKE_CURRENT_SOURCE_DIR}/layer/log_scanner.cc
 )
 target_include_directories(performance_layers_support_lib INTERFACE
@@ -70,9 +71,6 @@ target_link_libraries(performance_layers_support_lib INTERFACE
     absl::str_format
     absl::synchronization
     farmhash
-)
-target_compile_definitions(performance_layers_support_lib INTERFACE
-	STADIA_PERFORMANCE_LAYERS_NO_GLOG=1
 )
 
 # Layer targets.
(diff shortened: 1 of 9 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2020-2021 Google LLC
```
dnovillo: This is from #30 right?

layer/layer_utils.h
```diff
@@ -15,11 +15,47 @@
...
+namespace performancelayers {
+enum class LogMessageKind { INFO, WARNING, ERROR };
+
+// Helper class for printing log messages. Exposes a common logging macro,
+// SPL_LOG, independent of the underlying logging library. Sample use:
+//   SPL_LOG(WARNING) << "Cannot load file: " << my_file_path;
+//
+// Note that there is no need to add a newline character at the end -- this is
+// handled by the implementation.
+class MessageLogger {
```
dnovillo: nit: should this be in a `layers/log.h` file instead?  To prevent accumulating unrelated stuff in this generic file.

layer/runtime_layer.cc
```diff
@@ -12,14 +12,14 @@
...
 
-#include "runtime_layer_data.h"
-
 #include <cassert>
 #include <cstdint>
 #include <cstring>
 #include <functional>
 #include <string>
 
+#include "runtime_layer_data.h"
```
dnovillo, line 21: What does this include bring?  It does not seem to be used.

dnovillo (changes requested):

layer/runtime_layer.cc
kuhar: Clang-format reordered includes. `RuntimeLayerData` is used in almost every function.

layer/layer_utils.h
```diff
@@ -15,11 +15,4 @@
 #ifndef STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_UTILS_H_
 #define STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_UTILS_H_
```
kuhar, line 16: This header is empty now, but I'll layer macros here in a next PR. I didn't remove it entirely because of that.

layer/layer_utils.h
kuhar: Changed this to logging.h. Thanks for the suggestion.

dnovillo (approved):
Thanks!  LGTM.

PR #32, opened 2021-02-25
Automatically register all intercepted vulkan functions

kuhar:
This has been by far the largest source of bugs in performance layers.
It's very easy to forget to register a function manually, forget that
a function must be registered both as a device and instance override,
or defining an intercepting funciton with a slightly different
signature. These bugs are quite frustrating to debug, because they often
only show up with multiple layers running at the same time.

This PR implements an automatic function registration, similar to how
unit testing framework register test cases. We use a dedicated macro
to define all layer functions. This will make sure that the function
types match, and will add the function registration.

With all intercepted function registered globally, it is no longer
necessary to list all intercepted function in Get*ProcAddress entry
points.

Also move dispatch layer construction macros to the layer_utils.h
header.

Tested locally with all performance layers enabled on vkcube and a
couple of SaschaWillems examples.


sudonatalie (approved):

jaebaek (approved):
Looks good to me

PR #33, opened 2021-02-26
Implement InputBuffer class for reading files

kuhar:
Provide two concrete implementations:
1. Cross-platform file-based (fopen/fread/fclose).
2. Memory mapped implementation for unix.

Add InputBuffer tests.

This will be used by the cache sideload layer.

### DIFF
--- a/layer/input_buffer.cc
+++ b/layer/input_buffer.cc
@@ -1,0 +1,214 @@
+// Copyright 2021 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "input_buffer.h"
+
+#include <cassert>
+#include <cstdio>
+#include <vector>
+
+#include "absl/strings/str_cat.h"
+#include "logging.h"
+
+#if defined(__unix__)
+#include <fcntl.h>
+#include <sys/io.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <unistd.h>
+#endif
+
+namespace performancelayers {
+
+namespace {
+// Cross-platfrom input buffer implementation that reads file contents using
+// fopen/fread/fclose.
+class FileInputBufferImpl : public InputBuffer::InputBufferImplBase {
+ public:
+  static absl::StatusOr<std::unique_ptr<InputBuffer::InputBufferImplBase>>
+  Create(const std::string& path) {
+    struct FileCloser {
+      void operator()(FILE* file) {
+        if (file) fclose(file);
+      }
+    };
+    std::unique_ptr<FILE, FileCloser> handle(fopen(path.c_str(), "rb"));
+    if (!handle) {
+      return absl::UnavailableError(
+          absl::StrCat("Failed to fopen file for read: ", path));
+    }
+
+    fseek(handle.get(), 0, SEEK_END);
+    const auto file_size = ftell(handle.get());
+    fseek(handle.get(), 0, SEEK_SET);
+    if (file_size < 0) {
+      return absl::UnavailableError(
+          absl::StrCat("Failed to read file size: ", path));
+    }
+
+    const auto buffer_size = static_cast<size_t>(file_size);
+    FileInputBufferImpl res;
+    res.buffer_.resize(buffer_size);
+    if (fread(res.buffer_.data(), 1, buffer_size, handle.get()) !=
+        buffer_size) {
+      return absl::UnavailableError(
+          absl::StrCat("Failed to read file buffer contents: ", path));
+    }
+
+    return std::make_unique<FileInputBufferImpl>(std::move(res));
+  }
+
+  absl::Span<const uint8_t> GetBuffer() const override { return buffer_; }
+  ~FileInputBufferImpl() override = default;
+
+ private:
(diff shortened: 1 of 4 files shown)


layer/input_buffer.h
```diff
@@ -0,0 +1,74 @@
...
+  InputBuffer(std::unique_ptr<InputBufferImplBase> impl)
+      : concrete_impl_(std::move(impl)) {
+    assert(concrete_impl_);
+  }
+  std::unique_ptr<InputBufferImplBase> concrete_impl_ = nullptr;
+};
+
+}  // namespace performancelayers
+
+#endif  // STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_INPUT_BUFFER_
```
sudonatalie: nit: Should this be STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_INPUT_BUFFER_H_ ?

sudonatalie (approved):

layer/input_buffer.h
kuhar: Good catch, thanks

layer/input_buffer.cc
```diff
@@ -0,0 +1,209 @@
...
+        if (file) fclose(file);
+      }
+    };
+    std::unique_ptr<FILE, FileCloser> handle(fopen(path.c_str(), "rb"));
+    if (!handle) {
+      return absl::UnavailableError(
+          absl::StrCat("Failed to fopen file for read: ", path));
+    }
+
+    fseek(handle.get(), 0, SEEK_END);
```
jaebaek, line 53: I do not have experiences using `fseek`, `fopen`, `fclose`, ... on Windows. Does it work fine?

layer/input_buffer.cc
```diff
@@ -0,0 +1,209 @@
...
+          absl::StrCat("Failed to stat file: ", path));
+    }
+    size_t size = s.st_size;
+    if (size == 0) {
+      // We cannot mmap empty files, but we can return an empty buffer without
+      // mmaping the underlying file.
+      return std::make_unique<UnixMemMappedInputBufferImpl>(std::move(res));
+    }
+
+    void* data = mmap(0, size, PROT_READ, MAP_PRIVATE, res.file_descriptor_, 0);
```
jaebaek, line 108: I am confused .. if we use `PROT_READ`, we have to only read something from it. As far as I remember, writing sth on it causes a fault.

If I understand it correctly, why don't we use `const void* data` instead of `void* data`?
Likewise, I am not sure we can use it for `uint8_t* buffer_start_ = nullptr;` (not `const`).

layer/input_buffer.cc
```diff
@@ -0,0 +1,209 @@
...
+    file_descriptor_ = -1;
+    buffer_start_ = nullptr;
+    buffer_size_ = 0;
+  }
+
+  int file_descriptor_ = -1;
+  uint8_t* buffer_start_ = nullptr;
+  size_t buffer_size_ = 0;
+};
+#endif
```
jaebaek: nit: `#endif // defined(__unix__)`

layer/input_buffer.cc
```diff
@@ -0,0 +1,209 @@
...
+    }
+    case ImplementationKind::kMemMapped: {
+#if defined(__unix__)
+      auto mmap_input_buffer_or_err =
+          UnixMemMappedInputBufferImpl::Create(path);
+      if (mmap_input_buffer_or_err.ok())
+        return InputBuffer(std::move(*mmap_input_buffer_or_err));
+
+      return mmap_input_buffer_or_err.status();
+#else
```
jaebaek: `#else // defined(__unix__)`

layer/input_buffer.cc
```diff
@@ -0,0 +1,209 @@
...
+      return absl::UnavailableError(
+          "kMemMapped InputBuffer is implemented for this platform");
+#endif
```
jaebaek: Likewise

layer/input_buffer.cc
```diff
@@ -0,0 +1,209 @@
...
+    default:
+      assert(false && "Case not handled.");
+  }
+}
+
+absl::StatusOr<InputBuffer> InputBuffer::Create(const std::string& path) {
+#if defined(__unix__)
+  constexpr auto kPreferredPlatformImpl =
+      InputBuffer::ImplementationKind::kMemMapped;
+#else
```
jaebaek, line 206: ditto

layer/input_buffer.cc
```diff
@@ -0,0 +1,209 @@
...
+  constexpr auto kPreferredPlatformImpl =
+      InputBuffer::ImplementationKind::kFileRead;
+#endif
```
jaebaek: ditto

layer/input_buffer.cc
kuhar: These come from `<cstdio>`. All standard library functions should work on all supported platforms.

layer/input_buffer.cc
kuhar: Yes, this is read only. The only reason why I didn't use a const pointer is because `munmap` expects a `void*` parameter, and I didn't want to introduce `const_casts` for little benefit. This class only ever exposes the memory through a const span.

What do you think?

jaebaek: If we use it only insider this class, it would be fine. Could you please add a comment to explain this is a read only buffer?

kuhar: Done. Thanks for the suggestion.

layer/input_buffer.cc
```diff
@@ -0,0 +1,212 @@
...
+    if (buffer_start_) munmap(buffer_start_, buffer_size_);
+    if (file_descriptor_ != -1) close(file_descriptor_);
+
+    file_descriptor_ = -1;
+    buffer_start_ = nullptr;
+    buffer_size_ = 0;
+  }
+
+  int file_descriptor_ = -1;
+  uint8_t* buffer_start_ = nullptr;
```
jaebaek, line 166: Nit: please add a short comment for `uint8_t* buffer_start_ = nullptr;` as well.

jaebaek (approved):

PR #34, opened 2021-02-26
Add a layer for sideloading pipeline cache files

kuhar:
Pipeline cache sideload layer supplies pipeline caches to applications
that either do not use pipeline caches, or do not initialize them with
the intended initial data.

The pipeline cache file to load can be specified by setting the
`VK_PIPELINE_CACHE_SIDELOAD_FILE` environment variable.

The layer creates an implicit pipeline cache object for each device,
initialized with the specified file contents. This implicit cache then
gets merged into application pipeline caches (if any). The layer makes
sure that a valid pipeline cache handle is passed to every pipelien
creation.

This layer does not produce `.csv` log files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
 # Vulkan Performance Layers
 
-This project contains three Vulkan layers:
+This project contains 4 Vulkan layers:
 1. Compile time layer for measuring pipeline compilation times. The output log file location can be set with the `VK_COMPILE_TIME_LOG` environment variable.
 2. Runtime layer for measuring pipeline execution times. The output log file location can be set with the `VK_RUNTIME_LOG` environment variable.
 3. Frame time layer for measuring time between calls to vkQueuePresentKHR, in nanoseconds. This layer can also terminate the parent Vulkan application after a given number of frames, controlled by the `VK_FRAME_TIME_EXIT_AFTER_FRAME` environment variable. The output log file location can be set with the `VK_FRAME_TIME_LOG` environment variable. Benchmark start detection is controlled by the `VK_FRAME_TIME_BENCHMARK_WATCH_FILE` (which file to incrementally scan) and `VK_FRAME_TIME_BENCHMARK_START_STRING` (string that denotes benchmark start) environment variables.
+4. Pipeline cache sideloading layer for supplying pipeline caches to applications that either do not use pipeline caches, or do not initialize them with the intended initial data. The pipeline cache file to load can be specified by setting the `VK_PIPELINE_CACHE_SIDELOAD_FILE` environment variable. The layer creates an implicit pipeline cache object for each device, initialized with the specified file contents, which then gets merged into application pipeline caches (if any), and makes sure that a valid pipeline cache handle is passed to every pipeline creation. This layer does not produce `.csv` log files.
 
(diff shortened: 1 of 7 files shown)


layer/cache_sideload_layer.cc
```diff
@@ -0,0 +1,427 @@
...
+    return it->second;
+
+  return nullptr;
+}
+
+VkPipelineCache CacheSideloadLayerData::CreateImplicitDeviceCache(
+    VkDevice device, const VkAllocationCallbacks* alloc_callbacks,
+    absl::Span<const uint8_t> initial_data) {
+  absl::MutexLock lock(&device_to_implicit_cache_handle_lock_);
+  assert(device_to_implicit_cache_handle_.count(device) == 0 &&
```
jaebaek, line 67: If `absl::MutexLock` is an inter-process lock, `assert` failure of a process can put another process in the deadlock.
Is it ok?

sudonatalie (approved):

README.md
```diff
@@ -1,9 +1,10 @@
...
 3. Frame time layer for measuring time between calls to vkQueuePresentKHR, in nanoseconds. This layer can also terminate the parent Vulkan application after a given number of frames, controlled by the `VK_FRAME_TIME_EXIT_AFTER_FRAME` environment variable. The output log file location can be set with the `VK_FRAME_TIME_LOG` environment variable. Benchmark start detection is controlled by the `VK_FRAME_TIME_BENCHMARK_WATCH_FILE` (which file to incrementally scan) and `VK_FRAME_TIME_BENCHMARK_START_STRING` (string that denotes benchmark start) environment variables.
+4. Pipeline cache sideloading layer for supplying pipeline caches to applications that either do not use pipeline caches, or do not initialize them with the intended initial data. The pipeline cache file to load can be specified by setting the `VK_PIPELINE_CACHE_SIDELOAD_FILE` environment variable. The layer creates an implicit pipeline cache object for each device, initialized with the specified file contents, which then gets merged into application pipeline caches (if any), and makes sure that a valid pipeline cache handle is passed to every pipelien creation. This layer does not produce `.csv` log files.
```
sudonatalie: nit: s/pipelien/pipeline

kuhar: Thanks. I forgot to install a spellcheck plugin on my devkit. I also found a few other typos and fixed them.

jaebaek (commented):
I am curious if there is a way to add some unit tests, in particular, for functions in `layer/cache_sideload_layer.cc`.

layer/cache_sideload_layer.cc
kuhar: This is not an inter-process lock. It's local to the layer. Layer functions can be called concurrently, but the lock is not used by any other code. When the assertion fails, it will crash the whole application. This is the intended behavior.

kuhar:
@jaebaek 
> I am curious if there is a way to add some unit tests, in particular, for functions in layer/cache_sideload_layer.cc.

This is hard without a GPU VM. I thought of using a software ICD like swiftshader to test these layers using GitHub Action workers, but this seems like a non-trivial amount of work. W.r.t. this particular PR, I'm not sure that swiftshader even supports pipeline cache blobs.

jaebaek (approved):

PR #37, opened 2021-06-22
Use environment variables to load the layer under test.

nisse-work:
Set VK_LAYER_PATH and VK_INSTANCE_LAYERS, instead of modifying the
user's ~/.local/share/vulkan/implicit_layer.d.

Also use jq to extract needed values from the manifest json file,
instead of regexps.


kuhar (approved):
LGTM

PR #38, opened 2021-06-24
Add helper method LogLine

nisse-work:
Intended to be used by specific layers, without direct access to |out_| and |event_log_|.


kuhar (approved):
LGTM. Thanks for the cleanup!

PR #39, opened 2021-06-24
New helper class QueueToDeviceMap.

nisse-work:
Shared between the frame_time and runtime layers, without growing the
LayerData class. Intend to use also for the memory_usage layer. Also
fix runtime layer to use separate type-safe maps for VkCommandBuffer
and VkQueue, rather than a map keyed by void*.

### DIFF
--- a/layer/frame_time_layer.cc
+++ b/layer/frame_time_layer.cc
@@ -67,18 +67,17 @@ class FrameTimeLayerData : public performancelayers::LayerData {
 
   ~FrameTimeLayerData() override;
 
-  // Records the device that owns |queue|.
-  void SetDevice(VkQueue queue, VkDevice device) {
-    absl::MutexLock lock(&queue_to_device_lock_);
-    queue_to_device_.insert_or_assign(queue, device);
+  void GetDeviceQueue(VkDevice device, uint32_t queue_family_index,
+                      uint32_t queue_index, VkQueue* queue) {
+    queue_to_device_map_.GetDeviceQueue(this, device, queue_family_index, queue_index, queue);
   }
 
-  // Returns the device that owns |cmd_buffer|.
-  VkDevice GetDevice(VkQueue queue) const {
-    absl::MutexLock lock(&queue_to_device_lock_);
-    assert(queue_to_device_.count(queue) != 0);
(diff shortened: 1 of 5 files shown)


layer/frame_time_layer.cc
```diff
@@ -99,10 +98,7 @@ class FrameTimeLayerData : public performancelayers::LayerData {
   std::string benchmark_start_pattern_;
   std::optional<performancelayers::LogScanner> benchmark_log_scanner_;
 
-  mutable absl::Mutex queue_to_device_lock_;
-  // The map from a queue to the device that owns it.
-  absl::flat_hash_map<VkQueue, VkDevice> queue_to_device_
-      ABSL_GUARDED_BY(queue_to_device_lock_);
+  performancelayers::QueueToDeviceMap queue_to_device_map_;
```
nisse-work, line 104: There's some potential to make usage more concise if we inherit this class as a mixin, instead of adding it as a member. But I think multiple inheritance is discouraged, and I'm fine with a few single-line methods to explicitly forward work the the right member.

kuhar (commented):
I think this refactoring makes sense relative to the existing code, but if we could drop this map entirely if it's only used to get device proc addresses: https://github.com/googlestadia/performance-layers/issues/24

VkDevice, VkQueue, and VkCommandBuffer should share the same dispatch table: 
- https://github.com/LunarG/VulkanSamples/blob/16cb7667f5de831ac042c7fa9eb8c771260197fe/Sample-Programs/Hologram/generate-dispatch-table#L56.
- https://github.com/google/gf-layers/blob/master/src/gf_layers_layer_util/src/util.cc
- https://github.com/google/gf-layers/blob/98a9747b711838f49a296dada34ba9aa64346307/src/gf_layers_layer_util/include/gf_layers_layer_util/util.h#L148

kuhar:
Hi @nisse-work,

I opened a PR for what I suggested in the previous comment: https://github.com/googlestadia/performance-layers/pull/41. Could you see if it makes sense from the perspective of the memory usage layer?

nisse-work:
The changes in pr #41 makes this a lot simpler. Nice!

PR #40, opened 2021-06-24
Add a new layer, for measuring device memory usage

nisse-work:
Intercepts calls to vkAllocateMemory and vkFreeMemory, and emits a log line for each presented frame.

Maybe extended to in addition call VkPhysicalDeviceMemoryBudgetPropertiesEXT, and log this information too.


nisse-work:
Seems I've got wrong email attached to one of the commits. Will fix. Not sure it it's enough to just force-push to the branch this pr was made from, or if I'll have to abandon this one and make a new pr.

When this is ready for merging, I think it would make most sense to merge it as a single (squashed) change.

layer/VkLayer_stadia_memory_usage.json
```diff
@@ -0,0 +1,22 @@
+
+{
+  "file_format_version" : "1.0.0",
+  "layer" : {
+    "name": "VK_LAYER_STADIA_memory_usage",
+    "type": "GLOBAL",
+    "library_path": "libVkLayer_stadia_memory_usage.so",
+    "api_version": "1.1.74",
+    "implementation_version": "1",
+    "description": "Measures the memory usage of a vulken aplication.",
```
kuhar: typo in Vulkan

kuhar: Could you also clarify that this measures device/video memory usage?

layer/memory_usage_layer.cc
```diff
@@ -0,0 +1,335 @@
...
+// the layer data.
+SPL_MEMORY_USAGE_LAYER_FUNC(void, DestroyDevice,
+                            (VkDevice device,
+                             const VkAllocationCallbacks* allocator)) {
+  MemoryUsageLayerData* layer_data = GetLayerData();
+  // Remove memory allocation records for the device being destroyed.
+  layer_data->RecordDestroyDeviceMemory(device);
+  auto next_proc = layer_data->GetNextDeviceProcAddr(
+      device, &VkLayerDispatchTable::DestroyDevice);
+  (next_proc)(device, allocator);
```
kuhar: nit: we don't need parens around the function pointer, and I don't think it improves readability. This is done inconsistently in the other places in the other layers, and I think we  should eventually switch to the style without parens.

(also elsewhere in this file)

layer/memory_usage_layer.cc
```diff
@@ -0,0 +1,335 @@
+// Copyright 2020-2022 Google LLC
```
kuhar: I think this should be just 2021

kuhar (commented):
Overall, the implementation looks fine to me, I only found a few cosmetic issues.

Once the supporting PR lands, this should be rebased. This will also need some documentation (in README.md), but that can be a separate PR once this one is submitted.

layer/VkLayer_stadia_memory_usage.json
nisse-work: Done!

layer/memory_usage_layer.cc
nisse-work: Done.

layer/memory_usage_layer.cc
nisse-work: Ooops. Fixed.

layer/VkLayer_stadia_memory_usage.json
```diff
@@ -0,0 +1,22 @@
+
+{
+  "file_format_version" : "1.0.0",
+  "layer" : {
+    "name": "VK_LAYER_STADIA_memory_usage",
+    "type": "GLOBAL",
+    "library_path": "libVkLayer_stadia_memory_usage.so",
+    "api_version": "1.1.74",
+    "implementation_version": "1",
+    "description": "Measures the device memory usage of a vulkan aplication.",
```
kuhar: nit: could you capitalize 'vulkan'?

layer/memory_usage_layer.cc
```diff
@@ -0,0 +1,273 @@
+// Copyright 2020-2021 Google LLC
```
kuhar: AFAIK this is new code, so is 2020 necessary?

kuhar (commented):
Overall, looks good to me. I found just a few nits.

@nisse-work, could you also update the README with a short explanation of what the layer does and how to use it? And also add a sample CSV log file to the `sample_output/` directory.

Long term, I agree with the TODO saying that Device -- DeviceMemory -- Allocation should be a 2-level map, but I think the current code is fine for now.

layer/memory_usage_layer.cc
nisse-work: Parts were copied from the other layers, dated 2020-2021. But there were more copied code in earlier revisions, e.g., the now deleted logic for queue-to-device mapping. So changing to only 2021.

kuhar (approved):
LGTM. Thanks for the fixes.

PR #41, opened 2021-06-25
Allow to access dispatch tables with dispatchable keys

### LINKED ISSUE #24: Use dispatchable handles instead of tracking object relationships
This is a low priority refactoring task to simplify core layer code.
See the suggestion form @paulthomson: https://github.com/googlestadia/performance-layers/pull/18#issuecomment-698577374.

> In other words, instead of using VkInstance as a key (and the same idea applies for VkDevice), use the "dispatch table pointer" of the VkInstance, which you get using this function (please ignore the snake_case naming). And when you have a VkPhysicalDevice, use pretty much the same function to get the VkInstance dispatch table pointer, which you can then use to look-up the data you have for the VkInstance (which should include the VkInstance value itself).

> Another example of where this is somewhat nice is in a frame counter layer. We don't need to track VkQueue; we can go straight from the VkQueue to the DeviceData.

kuhar:
This is so that more handles can be used with `GetNext*ProcAddr`
functions. Removes some needless handle tracking across all layers.

Fixes: https://github.com/googlestadia/performance-layers/issues/24

### DIFF
--- a/layer/layer_data.h
+++ b/layer/layer_data.h
@@ -42,6 +42,87 @@ std::string CsvCat(Args&&... args) {
   return absl::StrJoin(std::forward_as_tuple(std::forward<Args>(args)...), ",");
 }
 
+// Base class for Vulkan dispatchable handle wrappers. Exposes the underlying
+// key to derived classes and implements basic operations like key comparisons
+// and hash.
+template <typename ConcreteKeyT>
+class DispatchableKeyBase {
+ public:
+  bool operator==(const ConcreteKeyT& other) const {
+    return key_ == other.key_;
+  }
+  bool operator!=(const ConcreteKeyT& other) const {
+    return key_ != other.key_;
+  }
+  size_t hash() const { return std::hash<void*>{}(key_); }
+
+ protected:
+  using BaseT = DispatchableKeyBase;
+  explicit DispatchableKeyBase(void** raw_handle) {
+    if (raw_handle) key_ = *raw_handle;
+  }
+  void* key_ = nullptr;
+};
+
+// InstanceKey provides the underlying instance key for dispatchable instance
+// handles: VkInstance and VkPhysicalDevice. Typically used with
+// `GetNextInstanceProcAddr`. Can be used as a hash map key.
+class InstanceKey : public DispatchableKeyBase<InstanceKey> {
+ public:
+  InstanceKey() : BaseT(nullptr) {}
+
+  explicit InstanceKey(VkInstance instance)
+      : BaseT(reinterpret_cast<void**>(instance)) {}
+  explicit InstanceKey(VkPhysicalDevice gpu)
+      : BaseT(reinterpret_cast<void**>(gpu)) {}
+
+  // Copyable and moveable.
+  InstanceKey(const InstanceKey&) = default;
+  InstanceKey(InstanceKey&&) = default;
+  InstanceKey& operator=(const InstanceKey&) = default;
+  InstanceKey& operator=(InstanceKey&&) = default;
+};
+
+// DeviceKey provides the underlying device key for dispatchable device
+// handles: VkDevice, VkQueue, and VkCommandBuffer. Typically used with
+// `GetNextDeviceProcAddr`. Can be used as a hash map key.
+class DeviceKey : public DispatchableKeyBase<DeviceKey> {
+ public:
+  DeviceKey() : BaseT(nullptr) {}
+
+  explicit DeviceKey(VkDevice device)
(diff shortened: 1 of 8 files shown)


kuhar:
@nisse-work please take a look

layer/layer_data.h
```diff
@@ -42,6 +42,87 @@ std::string CsvCat(Args&&... args) {
...
+// key to derived classes and implements basic operations like key comparisons
+// and hash.
+template <typename ConcreteKeyT>
+class DispatchableKeyBase {
+ public:
+  bool operator==(const ConcreteKeyT& other) const {
+    return key_ == other.key_;
+  }
+  bool operator!=(const ConcreteKeyT& other) const {
+    return key_ != other.key_;
```
vettoreldaniele, line 55:
```suggestion
    return !operator==(other);
```

layer/layer_data.h
```diff
@@ -62,67 +144,91 @@ class LayerData {
...
-  }
-
-  // Returns the instance associated with |gpu|, or a null handle if there is
-  // none.
-  VkInstance GetInstance(VkPhysicalDevice gpu) const {
-    absl::MutexLock lock(&gpu_instance_lock_);
-    if (auto it = gpu_instance_map_.find(gpu); it != gpu_instance_map_.end())
+  // Returns the instance associated with |instance_key|, or a null handle if
+  // there is none.
+  VkInstance GetInstance(InstanceKey instance_key) const {
```
vettoreldaniele, line 167: Since `AddInstance` and `RemoveInstance` take a `VkInstance`, can we have `GetInstance` also take a `VkInstance` and constructs the `InstanceKey` inside?

layer/layer_data.h
```diff
@@ -62,67 +144,91 @@ class LayerData {
...
+    absl::MutexLock lock(&instance_dispatch_lock_);
+    if (auto it = instance_keys_map_.find(instance_key);
+        it != instance_keys_map_.end())
       return it->second;
```
vettoreldaniele, line 171: nit: can you add `{}` brackets for the body of the if?

layer/layer_data.h
```diff
@@ -62,67 +144,91 @@ class LayerData {
...
+  }
+
+  // Returns the device associated with |device_key|, or a null handle if
+  // there is none.
+  VkDevice GetDevice(DeviceKey device_key) const {
     absl::MutexLock lock(&device_dispatch_lock_);
-    device_dispatch_map_.erase(device);
+    if (auto it = device_keys_map_.find(device_key);
+        it != device_keys_map_.end())
+      return it->second;
```
vettoreldaniele, line 203: nit: brackets

layer/layer_data.h
```diff
@@ -62,67 +144,91 @@ class LayerData {
...
   void RemoveDevice(VkDevice device) {
+    DeviceKey key(device);
+    absl::MutexLock lock(&device_dispatch_lock_);
+    device_dispatch_map_.erase(key);
+    device_keys_map_.erase(key);
+  }
+
+  // Returns the device associated with |device_key|, or a null handle if
+  // there is none.
+  VkDevice GetDevice(DeviceKey device_key) const {
```
vettoreldaniele, line 199: Same as `GetInstance`, can we have this take a `vkDevice` and construct the key inside?

layer/layer_data.h
kuhar: `GetInstance` is supposed to map from instance keys to VKInstances, passing a VkInstance will get you the same object back. I think it would make more sense to overload it to accept VkPhysicalDevice instead, but that circumvents the InstanceKey abstraction.
What do you think?

layer/layer_data.h
kuhar: Comparisons are not customizable by derived classes, so I don't think that this brings much values. I find the current version more readable, so I'll keep it.

layer/layer_data.h
kuhar: Same as above. It only makes sense to pass keys constructed from VkCommandBuffer and VkQueue here, passing VkDevice is an identity.

layer/layer_data.h
vettoreldaniele: Ah sorry you're right, it doesn't make sense to pass VkInstance. It's fine to leave it as it is.

layer/layer_data.h
vettoreldaniele: Yep, fine to leave as it is.

vettoreldaniele (approved):

kuhar:
V1: Fix cosmetic issues based on review comments.

PR #43, opened 2021-07-01
Delete left-over declaration of LayerData::EnumeratePhysicalDevices

nisse-work:
The implementation was deleted in https://github.com/googlestadia/performance-layers/pull/41


kuhar (approved):
Nice catch, thanks for fixing!

PR #44, opened 2021-07-05
Report cmake Vulkan configuration errors

kuhar:
... when `VULKAN_HEADERS_INSTALL_DIR` or `VULKAN_LOADER_GENERATED_DIR`
are set incorrectly.

Clarify build instructions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,21 +31,32 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
 add_subdirectory(third_party)
 
 if(DEFINED VULKAN_HEADERS_INSTALL_DIR)
+  if(NOT IS_ABSOLUTE "${VULKAN_HEADERS_INSTALL_DIR}")
+    message(FATAL_ERROR "VULKAN_HEADERS_INSTALL_DIR must be an absolute path")
+  endif()
   find_path(VulkanHeaders_INCLUDE_DIR
       NAMES vulkan/vk_layer.h
       HINTS ${VULKAN_HEADERS_INSTALL_DIR}/include
       NO_CMAKE_FIND_ROOT_PATH)
+  if(VulkanHeaders_INCLUDE_DIR MATCHES ".*_INCLUDE_DIR-NOTFOUND$")
+    message(FATAL_ERROR "Vulkan Headers not found. Set VULKAN_HEADERS_INSTALL_DIR to the Vulkan-Headers installation directory.")
+  endif()
 else()
   message(FATAL_ERROR "VULKAN_HEADERS_INSTALL_DIR must be specified!")
 endif()
-
 message(STATUS "VulkanHeaders ${VulkanHeaders_INCLUDE_DIR}")
 
 if(DEFINED VULKAN_LOADER_GENERATED_DIR)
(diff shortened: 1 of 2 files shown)


vettoreldaniele (approved):

CMakeLists.txt
```diff
@@ -31,21 +31,32 @@ set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC")
...
 
 if(DEFINED VULKAN_HEADERS_INSTALL_DIR)
+  if(NOT IS_ABSOLUTE "${VULKAN_HEADERS_INSTALL_DIR}")
+    message(FATAL_ERROR "VULKAN_HEADERS_INSTALL_DIR must be an absolute path")
+  endif()
   find_path(VulkanHeaders_INCLUDE_DIR
       NAMES vulkan/vk_layer.h
       HINTS ${VULKAN_HEADERS_INSTALL_DIR}/include
       NO_CMAKE_FIND_ROOT_PATH)
+  if(VulkanHeaders_INCLUDE_DIR MATCHES ".*_INCLUDE_DIR-NOTFOUND$")
```
nisse-work, line 41: If I remember correctly, incorrect VULKAN_HEADERS_INSTALL_DIR did not result in a _INCLUDE_DIR_NOT_FOUND, but it fell back to using /usr/local/include (since I had vulkan dev debian packages installed, but apparently an older version then required). I don't knowmuch about how cmake works, but maybe the HINTS above should be changed to something that requires that the include file is located in the given location?

kuhar: I tested this in the same configuration (i.e., with the debian package installed), and it didn't default to the system location.
>  I don't knowmuch about how cmake works, but maybe the HINTS above should be changed to something that requires that the include file is located in the given location?

This is what I thought `NO_CMAKE_FIND_ROOT_PATH` was supposed do. Now going over `find_path` again (https://cmake.org/cmake/help/latest/command/find_path.html), I can see some other ones that might apply:
1. `NO_DEFAULT_PATH`
2. `NO_CMAKE_FIND_ROOT_PATH`
3. `NO_PACKAGE_ROOT_PATH`
4. `NO_CMAKE_PATH`
5. `NO_CMAKE_ENVIRONMENT_PATH`
6. `NO_CMAKE_SYSTEM_PATH`

Perhaps there's a better function to use instead of `find_path`. Or as a workaround, we could check that the found path is actually an extension of `VULKAN_HEADERS_INSTALL_DIR`. What do you think?

nisse-work: I try the following, with a new, empty, build directory:

```
mkdir build-test
cd build-test
cmake .. \
-DVULKAN_LOADER_GENERATED_DIR=/home/nisse/hack/performance-layers/third_party/Vulkan-Loader/loader/generated \ 
-DVULKAN_HEADERS_INSTALL_DIR=/usr/local/does-not-exist
```

I get 
```
-- VulkanHeaders /usr/include
-- VulkanLoaderGenerated /home/nisse/hack/performance-layers/third_party/Vulkan-Loader/loader/generated
```

I would have expected configuration to fail. Instead, it succeeds, and falls back to /usr/include for the vulkan headers. My cmake is version 3.18.4.

kuhar: Thanks, I can reproduce it after installing the `vulkan-headers` apt package. I opened a PR with a fix: https://github.com/googlestadia/performance-layers/pull/49.

PR #48, opened 2021-07-14
Record delay between shader module creation and first use

kuhar:
... in pipeline creation.
    
The delay info is only logged to the even log.

### DIFF
--- a/layer/layer_data.h
+++ b/layer/layer_data.h
@@ -314,6 +314,9 @@ class LayerData {
   void LogEventOnly(std::string_view event_type,
                     std::string_view extra_content = "") const;
 
+  // Returns a string representation of |hash|.
+  static std::string ShaderHashToString(uint64_t hash);
+
   // Returns a string identifier of |pipeline|.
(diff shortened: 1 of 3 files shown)


kuhar:
Only the second commit is part of this PR. The first one is submitted for review separately: https://github.com/googlestadia/performance-layers/pull/47.

layer/layer_data.h
```diff
@@ -314,6 +314,9 @@ class LayerData {
   void LogEventOnly(std::string_view event_type,
                     std::string_view extra_content = "") const;
 
+  // Returns a string representation of |shader_hash|.
```
vettoreldaniele: Comment should say `|hash|`

vettoreldaniele (approved):
Thanks Jakub. I only reviewed the second commit as you mentioned.

kuhar:
Thanks for catching the outdated comment, @vettoreldaniele. Could you also take a look at #47?

kuhar:
V2: Rebased.

PR #49, opened 2021-07-15
Refine Vulkan header search

kuhar:
Do not use default paths, consider only the user-specified directories.
This is based on
https://cmake.org/cmake/help/latest/command/find_path.html#:~:text=if%20no_default_path%20is%20specified.

Test: Install the `vulkan-headers` apt package and provide a wrong
absolute path. Make sure than an error is emitted.


kuhar:
@nisse-work PTAL

nisse-work:
Thanks, looks reasonable to me.

dnovillo (approved):

PR #50, opened 2021-07-15
Measure shader module creation times

kuhar:
Make sure that the same time markers are used for both shader module
creation duration and shader use delay.


kuhar:
V1: Undo accidental changes to includes

vettoreldaniele (approved):

PR #55, opened 2021-10-04
Track shader module creation end in first use slack

kuhar:
... instead of shader module creation start. This is because it seems
more useful to know how much slack time there is after
`vkCreateShaderModule` returns as applications most likely block on
this API call. Otherwise making the Vulkan implementation slower in
shader module creation would increate the reported delay, which seemed
counterintuitive to me.

Also rename `shader_module_first_use_delay_ns` to
`shader_module_first_use_slack_ns`. This doesn't require any script
changes.

### DIFF
--- a/layer/compile_time_layer.cc
+++ b/layer/compile_time_layer.cc
@@ -41,53 +41,53 @@ class CompileTimeLayerData : public performancelayers::LayerData {
     LogEventOnly("compile_time_layer_init");
   }
 
-  // Used to track the delay between shader module creation and its first use in
+  // Used to track the slack between shader module creation and its first use in
   // pipeline creation.
-  struct ShaderModuleDelay {
-    std::optional<absl::Time> creation_time = std::nullopt;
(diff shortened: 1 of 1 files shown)


layer/compile_time_layer.cc
```diff
@@ -44,43 +44,43 @@ class CompileTimeLayerData : public performancelayers::LayerData {
   // Used to track the delay between shader module creation and its first use in
   // pipeline creation.
   struct ShaderModuleDelay {
```
vettoreldaniele: Can you update the comment and the struct name to mention this is now slack time?

vettoreldaniele (changes requested):

layer/compile_time_layer.cc
kuhar: Good catch,  I changed both the struct name and comments.

vettoreldaniele (approved):
Thanks!

PR #61, opened 2022-08-29
remove an unneeded import of a legacy header file

lexi-nadia:
This file doesn't exist on newer platforms like aarch64.


kuhar:
Thanks @lexi-nadia! Merged.

PR #63, opened 2022-09-13
Erase destroyed shader modules from LayerData. 

### LINKED ISSUE #51: Handle vkDestroyShaderModule
`LayerData` should handle `vkDestroyShaderModule` so that we don't accidentally use pointers to objects whose lifetime ended.
https://www.khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkDestroyShaderModule.html

miladHakimi:
Previously, when DestroyShaderModule was called, the module was removed
but its record remained in the LayerData. Since the hash function uses
the pointer and the size of the shader module to create a key, another
shader module that is allocated in the same address and has the same
size could potentially cause a bug. Because its corresponding key will
be the same as the key that already exist in the LayerData.

Remove a shader module from the LayerData when DestroyShaderModule is
called. This makes sure that we don't keep record of the destroyed
objects.

Add a hook on DestroyShaderModule function in the compile_time_layer and
runtime_layer.

Fixes: https://github.com/googlestadia/performance-layers/issues/51

### DIFF
--- a/layer/compile_time_layer.cc
+++ b/layer/compile_time_layer.cc
@@ -242,7 +242,15 @@ SPL_COMPILE_TIME_LAYER_FUNC(VkResult, CreateShaderModule,
   return res.result;
 }
 
-// Override for vkDestroyDevice.  Removes the dispatch table for the device from
+// Override for vkDestroyShaderModule. Erases the shader module from the layer
+// data.
+SPL_COMPILE_TIME_LAYER_FUNC(void, DestroyShaderModule,
+                            (VkDevice device, VkShaderModule shader_module,
+                             const VkAllocationCallbacks* allocator)) {
+  return GetLayerData()->DestroyShaderModule(device, shader_module, allocator);
+}
+
+// Override for vkDestroyDevice. Removes the dispatch table for the device from
 // the layer data.
 SPL_COMPILE_TIME_LAYER_FUNC(void, DestroyDevice,
                             (VkDevice device,
@@ -254,7 +262,7 @@ SPL_COMPILE_TIME_LAYER_FUNC(void, DestroyDevice,
   next_proc(device, allocator);
 }
 
-// Override for vkCreateDevice.  Builds the dispatch table for the new device
+// Override for vkCreateDevice. Builds the dispatch table for the new device
 // and add it to the layer data.
 SPL_COMPILE_TIME_LAYER_FUNC(VkResult, CreateDevice,
                             (VkPhysicalDevice physical_device,
@@ -270,6 +278,8 @@ SPL_COMPILE_TIME_LAYER_FUNC(VkResult, CreateDevice,
     SPL_DISPATCH_DEVICE_FUNC(CreateComputePipelines);
     SPL_DISPATCH_DEVICE_FUNC(CreateGraphicsPipelines);
     SPL_DISPATCH_DEVICE_FUNC(CreateShaderModule);
+    SPL_DISPATCH_DEVICE_FUNC(DestroyShaderModule);
+
     return dispatch_table;
   };
 
--- a/layer/layer_data.cc
+++ b/layer/layer_data.cc
@@ -269,4 +269,13 @@ LayerData::ShaderModuleCreateResult LayerData::CreateShaderModule(
       HashShader(*shader_module, create_info->pCode, create_info->codeSize);
   return {result, hash, start, end};
 }
+
+void LayerData::DestroyShaderModule(VkDevice device,
+                                    VkShaderModule shader_module,
+                                    const VkAllocationCallbacks* allocator) {
+  auto next_proc =
+      GetNextDeviceProcAddr(device, &VkLayerDispatchTable::DestroyShaderModule);
+  EraseShader(shader_module);
+  next_proc(device, shader_module, allocator);
+}
 }  // namespace performancelayers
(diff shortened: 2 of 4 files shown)


layer/compile_time_layer.cc
```diff
@@ -242,7 +242,17 @@ SPL_COMPILE_TIME_LAYER_FUNC(VkResult, CreateShaderModule,
...
 }
 
-// Override for vkDestroyDevice.  Removes the dispatch table for the device from
+// Override for vkDestroyShaderModule. Erases the shader module from
+// the layer data.
+SPL_COMPILE_TIME_LAYER_FUNC(void, DestroyShaderModule,
+                            (VkDevice device,
+                             VkShaderModule shader_module,
+                             const VkAllocationCallbacks* allocator)
+                             ) {
```
kuhar: Can you run clang-format on your patch? Either in your editor (format selection) or via CLI:
` git diff main -U0 --no-color | clang-format-diff -p1 -i`

layer/layer_data.cc
```diff
@@ -269,4 +269,16 @@ LayerData::ShaderModuleCreateResult LayerData::CreateShaderModule(
...
 }
+
+void LayerData::DestroyShaderModule(
+      VkDevice device, VkShaderModule shader_module,
+      const VkAllocationCallbacks* allocator) {
+  auto next_proc =
+      GetNextDeviceProcAddr(device, &VkLayerDispatchTable::DestroyShaderModule);
+  absl::Time start = absl::Now();
+  next_proc(device, shader_module, allocator);
+  absl::Time end = absl::Now();
```
kuhar: I don't think we have any immediate use case for destruction time. Could we remove this until we find one?

layer/layer_data.h
```diff
@@ -240,6 +240,18 @@ class LayerData {
     return proc_addr;
   }
 
+  // Removes a previously created shader module from the LayerData
```
kuhar: nit: missing `.` at the end. Also, could you re-flow the comment? IE put it all one a single line and let clang-format figure out how to break it to fit the column limit.

layer/layer_data.h
```diff
@@ -240,6 +240,18 @@ class LayerData {
...
+  // This is called while destroying the shader module. Returs the hash
+  // value of the Shader module.
+  uint64_t EraseShader(VkShaderModule shader_module) {
+    uint64_t hash_value = GetShaderHash(shader_module);
+    absl::MutexLock lock(&shader_hash_lock_);
+    auto shader_iter = shader_hash_map_.find(shader_module);
+    assert(shader_iter != shader_hash_map_.end());
```
kuhar: I think this will generate unused variable warnings on non-debug build configurations. You should be able to check for presence/absence with `.count` or `.contains`.

layer/layer_data.h
```diff
@@ -353,6 +365,10 @@ class LayerData {
       VkDevice device, const VkShaderModuleCreateInfo* create_info,
       const VkAllocationCallbacks* allocator, VkShaderModule* shader_module);
 
+  void DestroyShaderModule(
```
kuhar: Could you add documentation for this function? Something brief like by the functions above.

layer/runtime_layer.cc
```diff
@@ -337,6 +337,17 @@ SPL_RUNTIME_LAYER_FUNC(VkResult, CreateShaderModule,
       .result;
 }
 
+// Override for vkDestroyShaderModule. Erases the shader module from
+// the layer data.
```
kuhar, line 341: nit: reflow this comment please

kuhar (changes requested):
Thanks for taking care of this issue. I have a few suggestions, mostly cosmetic/stylistic issues.

Could you update the PR title so that it fits the blank in this sentence: 'When applied, this commit will __' ? See https://cbea.ms/git-commit/#imperative. (The existing title kinda does it, but I think we could improve it with something like 'Handle shader module destruction', 'Erase destroyed shader modules', 'Handle vkDestroyShaderModule in blah', etc.)

Could you also add the link to the issue at the bottom? The convention is:
```
Fixes: <issue-url>
```

layer/layer_data.cc
```diff
@@ -269,4 +269,13 @@ LayerData::ShaderModuleCreateResult LayerData::CreateShaderModule(
...
   return {result, hash, start, end};
 }
+
+void LayerData::DestroyShaderModule(VkDevice device,
+                                    VkShaderModule shader_module,
+                                    const VkAllocationCallbacks* allocator) {
+  auto next_proc =
+      GetNextDeviceProcAddr(device, &VkLayerDispatchTable::DestroyShaderModule);
+  next_proc(device, shader_module, allocator);
+  EraseShader(shader_module);
```
kuhar, line 278: I wonder if the `shader_module` handle is still valid at this point. If the memory gets deallocated at the end of the chain, then this may exhibit undefined behavior (using a pointer to an object whose lifetime has ended). I think this article explains this, based on a very quick read: https://blog.panicsoftware.com/objects-their-lifetimes-and-pointers/#:~:text=std%3A%3Alaunder%20and,in%20C%2B%2B17..

kuhar:
Could you also update the PR title / description on the website? When we squash commits uses these instead of actual commit messages.

kuhar (approved):
LGTM. I'll leave this open for a bit to see if @cassiebeckley has some suggestions.

kuhar:
> Could you also add the link to the issue at the bottom? The convention is:

@miladHakimi Could you also add this to the PR description?

cassiebeckley (approved):
LGTM as well

PR #64, opened 2022-09-13
Refactor shader_hash_map_ variable for more readability

miladHakimi:
`shader_hash_map_` is a variable that maps shader module to shader hashes. 
The name might be confusing because it can be misinterpreted as a variable that
holds the hash of the shader code. `Shader_to_code_hash` is a more descriptive
word and it's more aligned with this functionality.


kuhar (approved):
LGTM. Thanks for improving this!

PR #65, opened 2022-09-16
Refactor check-layers.yml for better resource utilization.

miladHakimi:
The current CI is checking for every combination of config, compiler, and generator. The checks contain redundancy without introducing more coverage. To reduce the redundant checks, only a subset of the combinations should be executed.

The building for both ninja and Makefiles is redundant, so we mainly build for ninja, but have a single check for Makefiles.

Keep both gcc and clang to exploit their different optimizations and diagnostic features, and keep the checks independent of compiler-specific behaviors. No need to compile with both debug and release flags for both compilers.

Set the default config, compiler, and generator in the job's matrix. The include filed expands the matrix by adding `Makefiles, gcc, Release` and `Ninja, gcc, Debug` configurations.

### DIFF
--- a/.github/workflows/check-layers.yml
+++ b/.github/workflows/check-layers.yml
@@ -7,14 +7,20 @@ on:
 jobs:
   build:
     name: Check (${{ matrix.config }}, ${{ matrix.compiler }}, ${{ matrix.generator }})
-    runs-on: ${{ matrix.host-os }}
+    runs-on: "ubuntu-latest"
     strategy:
       fail-fast: false
       matrix:
-        host-os:    ["ubuntu-latest"]
         config:     [Debug, Release]
-        compiler:   [clang, gcc]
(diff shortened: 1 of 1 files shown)


kuhar (approved):
Awesome! I didn't know about the `include` mechanism, that makes it much cleaner.

.github/workflows/check-layers.yml
```diff
@@ -13,8 +13,17 @@ jobs:
       matrix:
         host-os:    ["ubuntu-latest"]
```
vettoreldaniele: Since `ubuntu-latest` is the only platform used, you can remove it from the matrix and just use
```
runs-on: "ubuntu-latest"
```

kuhar:
@vettoreldaniele Could you take a look and check if this looks fine to you too. If yes, feel free to merge.

PR #66, opened 2022-09-21
Update dockerfile to install Vulkan SDK.

miladHakimi:
Performance layers are not tested on a Vulkan application. Here, the Vulkan application is `vkcube`. To test the layers on a docker, the Vulkan application must be executed in headless mode. `xvfb` allows us to run in headless mode by performing all the graphical operations in memory instead of showing them on a display.

### DIFF
--- a/docker/build.Dockerfile
+++ b/docker/build.Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2020-2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


docker/build.Dockerfile
```diff
@@ -1,3 +1,9 @@
+FROM ubuntu:20.04
+
+
+ARG CONFIG
+ARG COMPILER
+ARG GENERATOR
 # Copyright 2020 Google LLC
```
kuhar: Could you also update the year here? `2020-2022`

docker/build.Dockerfile
```diff
@@ -76,3 +88,12 @@ RUN  CXX_COMPILER="g++" \
     && cmake --build . --target check \
     && cmake --build . --target install
 
+# Install Vulkan SDK
+RUN wget -qO - https://packages.lunarg.com/lunarg-signing-key-pub.asc | apt-key add - \
+    && wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.3.216-bionic.list https://packages.lunarg.com/vulkan/1.3.216/lunarg-vulkan-1.3.216-bionic.list \
+    && apt-get update \
+    && apt-get install -yqq --no-install-recommends \
+       vulkan-sdk x11-xserver-utils libvulkan-dev libvulkan1 xvfb mesa-vulkan-drivers
```
kuhar: I think it'd be better to install everything in the first `RUN` command. This way we can fail early when the installation fails for whatever reason.

kuhar: Also, doesn't ubuntu come with some default packages for vulkan? Maybe we could skip the step where we add the key and download the list from lunarg?

kuhar (changes requested):

docker/build.Dockerfile
```diff
@@ -1,3 +1,9 @@
+FROM ubuntu:20.04
+
+
+ARG CONFIG
+ARG COMPILER
+ARG GENERATOR
```
kuhar: Copyright notice should be at the very top of the file

docker/build.Dockerfile
miladHakimi: Ubuntu doesn't come with default packages for Vulkan. If I remove the step containing adding the key and downloading the list, it will be unable to locate vulkan-sdk and the installation fails.

kuhar (approved):
LGTM

cassiebeckley (approved):
LGTM

PR #67, opened 2022-09-22
Run vkcube with performance layers enabled.

miladHakimi:
After the build was done, the layers were not tested on an actual Vulkan
application.
All the layers are now enabled by getting included in `VK_INSTANCE_LAYERS`. 
By running `vkcube` we check if an application crashes with layers enabled.

### DIFF
--- a/docker/build.Dockerfile
+++ b/docker/build.Dockerfile
@@ -1,4 +1,4 @@
-# Copyright 2020 Google LLC
+# Copyright 2020-2022 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


docker/build.Dockerfile
```diff
@@ -1,4 +1,4 @@
-# Copyright 2020-2022 Google LLC
+# Copyright 2020-2022-2022 Google LLC
```
kuhar: The year range already ends with 2022, no need for this modification

docker/build.Dockerfile
```diff
@@ -87,5 +87,12 @@ RUN  CXX_COMPILER="g++" \
...
-RUN xvfb-run vkcube --c 1000 && echo Done!
+# Enable perfomance layers and test with `vkcube`
+RUN export LD_LIBRARY_PATH=/performance-layers/build:$LD_LIBRARY_PATH \ 
+    && export VK_INSTANCE_LAYERS=VK_LAYER_STADIA_pipeline_compile_time \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_runtime \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_cache_sideload \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_memory_usage \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_frame_time \
+    && export VK_LAYER_PATH=/performance-layers/layer \
+    && xvfb-run vkcube --c 1000 && echo Done!
```
kuhar: nit: missing newline at the end of the file

docker/build.Dockerfile
```diff
@@ -87,5 +87,12 @@ RUN  CXX_COMPILER="g++" \
     && cmake --build . --target check \
     && cmake --build . --target install
 
-# Test Vulkan SDK
-RUN xvfb-run vkcube --c 1000 && echo Done!
+# Enable perfomance layers and test with `vkcube`
```
kuhar: nit: missing `.` at the end of the sentence

docker/build.Dockerfile
kuhar: 1000 frames take ~17s to render on my laptop, should we lower it to 100?

kuhar (changes requested):

docker/build.Dockerfile
```diff
@@ -87,5 +87,12 @@ RUN  CXX_COMPILER="g++" \
...
+RUN export LD_LIBRARY_PATH=/performance-layers/build:$LD_LIBRARY_PATH \ 
+    && export VK_INSTANCE_LAYERS=VK_LAYER_STADIA_pipeline_compile_time \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_runtime \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_cache_sideload \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_memory_usage \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_frame_time \
```
kuhar: These two names may be wrong: not all layers contain `pipeline_` in their name. Layer `.json` files define these names.

docker/build.Dockerfile
kuhar: Or does it take less time with xvfb than with a physical dispaly?

miladHakimi: It took about 2s on my machine both with physical display and xvfb. I'll bring it down to 100.

kuhar (approved):
LGTM, I see the logs I was expecting in the most recent CI runs.

PR #68, opened 2022-09-23
Test if enabled layers are populating their log files.

miladHakimi:
Setting the environment variables and running an application does not necessarily guarantee that the layers are enabled. Because the program can still run even if the environment variables are set incorrectly. There are many ways that allow us to check whether the layers are enabled or not. 
We can run `vulkaninfo` and look at the `layers` subsection. We can also use `grep <name-of-the-layer>` to  check whether or not they exist in `vulkaninfo`.
Moreover, when a layer is enabled, it produces logs and stores them in its log file which can be specified by setting its corresponding environment variable. By inspecting these log files we can determine the execution of the layer.

Since, in the next steps, we'd like to make sure that the logs are consistent with each other (and the input arguments), we've decided to use the second approach (inspecting the log files). The log files contain a header row followed by the data row(s). Here, we only examine the generation of data rather than the content of data. Hence, we only need to make sure that the log files contain a data row (has at least two rows).

### DIFF
--- a/docker/build.Dockerfile
+++ b/docker/build.Dockerfile
@@ -23,6 +23,8 @@
 
 FROM ubuntu:20.04
 
+SHELL ["/bin/bash", "-c"]
+
 ARG CONFIG
 ARG COMPILER
 ARG GENERATOR
@@ -87,12 +89,6 @@ RUN  CXX_COMPILER="g++" \
     && cmake --build . --target check \
     && cmake --build . --target install
 
-# Enable perfomance layers and test with `vkcube`.
-RUN export LD_LIBRARY_PATH=/performance-layers/build:$LD_LIBRARY_PATH \
-    && export VK_INSTANCE_LAYERS=VK_LAYER_STADIA_pipeline_compile_time \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_runtime \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_cache_sideload \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_memory_usage \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_frame_time \
-    && export VK_LAYER_PATH=/performance-layers/layer \
-    && xvfb-run vkcube --c 100 && echo Done!
+# Enable and test the performance layers.
+WORKDIR /performance-layers/docker
+RUN ./test_performance_layers.sh /performance-layers/build
--- a/docker/test_performance_layers.sh
+++ b/docker/test_performance_layers.sh
@@ -1,0 +1,57 @@
+#!/usr/bin/env bash
+#
+# Enables the performance layers by setting the evironment variables.
+# Executes vkcube and makes sure it runs without fault when performance layers 
+# are enabled.
+# Tests if each layer writes to its corresponding output file. 
+
+set -euo pipefail
+
+# Create a temp output directory for log files. 
+readonly OUTPUT_DIR="$(mktemp -d)"
+readonly INSTALL_DIR="$1"
+readonly LAYER_DIR="${INSTALL_DIR}/run/layer"
+
+declare -a output_files
+output_files=("compile_time.csv" "run_time.csv" "memory_usage.csv" 
+              "frame_time.csv" "events.log")
+
+#######################################
+# Checks if the layers write data in their specified log files.
+# If data does not exist in the file, writes an error message to stdout and
+# exits.
+# Globals:
+#   OUTPUT_DIR
+# Arguments:
+#   layer's log file name
+#######################################
+check_layer_log() {
+  local file_name="$1"
+
+  if [[ $(wc -l <"${OUTPUT_DIR}"/"${file_name}") -lt 2 ]]; then
+    echo "Error: ${file_name} has not been populated properly." 
+    exit 1
+  fi
+}
+
+export LD_LIBRARY_PATH="${INSTALL_DIR}":"${LD_LIBRARY_PATH-}"
+export VK_INSTANCE_LAYERS=VK_LAYER_STADIA_pipeline_compile_time
+export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_runtime
+export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_frame_time
+export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_memory_usage
+export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_cache_sideload
+export VK_LAYER_PATH="${LAYER_DIR}"
+export VK_COMPILE_TIME_LOG="${OUTPUT_DIR}"/compile_time.csv
+export VK_RUNTIME_LOG="${OUTPUT_DIR}"/run_time.csv
+export VK_FRAME_TIME_LOG="${OUTPUT_DIR}"/frame_time.csv
+export VK_MEMORY_USAGE_LOG="${OUTPUT_DIR}"/memory_usage.csv
+export VK_PERFORMANCE_LAYERS_EVENT_LOG_FILE="${OUTPUT_DIR}"/events.log
+
+# Test if an application can run with performance layers enabled.
+xvfb-run vkcube --c 100
+echo "Run is finished successfully!"
+
+# Check if each enabled layer produced its log file.
+for file in "${output_files[@]}"; do
+  check_layer_log "${file}"
+done


docker/build.Dockerfile
```diff
@@ -87,12 +87,37 @@ RUN  CXX_COMPILER="g++" \
...
     && export VK_INSTANCE_LAYERS=VK_LAYER_STADIA_pipeline_compile_time \
     && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_runtime \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_cache_sideload \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_memory_usage \
     && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_frame_time \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_memory_usage \
+    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_cache_sideload \
     && export VK_LAYER_PATH=/performance-layers/layer \
-    && xvfb-run vkcube --c 100 && echo Done!
+    && export VK_COMPILE_TIME_LOG=/performance-layers/build/compile_time.csv \
```
kuhar: In case the build succeeds and someone wants to extract the artefacts, I think it would be better not to write anything to the build directory. We can instead create a new temporary output directory with `mktmep -d`.

docker/build.Dockerfile
```diff
@@ -87,12 +87,37 @@ RUN  CXX_COMPILER="g++" \
...
+    && export VK_RUNTIME_LOG=/performance-layers/build/run_time.csv \ 
+    && export VK_FRAME_TIME_LOG=/performance-layers/build/frame_time.csv \
+    && export VK_MEMORY_USAGE_LOG=/performance-layers/build/memory_usage.csv \
+    && export VK_PIPELINE_CACHE_SIDELOAD_FILE=/performance-layers/build/cache_sideload_file.csv \
+    && export VK_PERFORMANCE_LAYERS_EVENT_LOG_FILE=/performance-layers/build/events.log \
+    && touch /performance-layers/build/cache_sideload_file.csv \
+    && xvfb-run vkcube --c 100 && echo Run is completed! 
+
+# Check if each enabled layer is writing in its corresponding log file.
+RUN if [ `wc -l <compile_time.csv` -lt 2 ]; then \
```
kuhar: Since the check appears to be the same for each of the layers, can we make it a function?

kuhar: Also, prefer subshells in the form of `$(your cmd)` over backticks. It comes with fewer surprised and allows for nesting. See http://mywiki.wooledge.org/BashFAQ/082

kuhar (changes requested):
This looks great, but could we move these two run tiles to a shell script? I think this will be easier to work with if we need to add more complicated logic in the future, and would allow us to test the layers outside of docker if we need to. The path to the install directory would be the input.

docker/build.Dockerfile
miladHakimi: That's a great idea.

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
...
+export VK_LAYER_PATH="${LAYER_DIR}"
+export VK_COMPILE_TIME_LOG="${OUTPUT_DIR}"/compile_time.csv
+export VK_RUNTIME_LOG="${OUTPUT_DIR}"/run_time.csv
+export VK_FRAME_TIME_LOG="${OUTPUT_DIR}"/frame_time.csv
+export VK_MEMORY_USAGE_LOG="${OUTPUT_DIR}"/memory_usage.csv
+export VK_PIPELINE_CACHE_SIDELOAD_FILE="${OUTPUT_DIR}"/cache_sideload_file.csv
+export VK_PERFORMANCE_LAYERS_EVENT_LOG_FILE="${OUTPUT_DIR}"/events.log
+
+# Create the input file for cache sideload pipeline.
+touch "${OUTPUT_DIR}"/cache_sideload_file.csv
```
kuhar: This is won't be a a valid pipeline cache file. According to the Vulkan specs, pipeline cache blob has to contain a header. So if we really wanted to pass some input, we could create something like a 256-byte binary file filled with all zeros, e.g., using `dd`. But I think it would also be fine not to give it any input, because the driver will reject it anyway, because both the header and the content are invalid.

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+#
+# Enable the performance layers by setting the evironment variables.
+# Run vkcube and make sure it runs without fault when performance layers are 
+# enabled.
+# Test if each layer write to its corresponding output file. 
+
+set -e
```
kuhar: Maybe `set -euo pipefail`?

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
...
+
+#######################################
+# Check if the layers write data in their specified log files.
+# Globals:
+#   OUTPUT_DIR
+# Arguments:
+#   layer's log file name
+# Outputs:
+#   If data does not exist in the file, write an error message to stdout and
+#   exit.
```
kuhar: I would not consider this an output. Consider moving this sentence to the description before `Globals:`.

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
...
+# Create a temp output directory for log files. 
+readonly OUTPUT_DIR="$(mktemp -d)"
+readonly INSTALL_DIR="$1"
+readonly LAYER_DIR="$2"
+
+declare -a output_files
+output_files=("compile_time.csv" "run_time.csv" "memory_usage.csv" "frame_time.csv")
+
+#######################################
+# Check if the layers write data in their specified log files.
```
kuhar: Excellent, this documentation is very useful and clear!

nit: `s/Check/Checks/`

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
...
+
+set -e
+
+# Create a temp output directory for log files. 
+readonly OUTPUT_DIR="$(mktemp -d)"
+readonly INSTALL_DIR="$1"
+readonly LAYER_DIR="$2"
+
+declare -a output_files
+output_files=("compile_time.csv" "run_time.csv" "memory_usage.csv" "frame_time.csv")
```
kuhar: What about `events.log`?

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
...
+export VK_FRAME_TIME_LOG="${OUTPUT_DIR}"/frame_time.csv
+export VK_MEMORY_USAGE_LOG="${OUTPUT_DIR}"/memory_usage.csv
+export VK_PIPELINE_CACHE_SIDELOAD_FILE="${OUTPUT_DIR}"/cache_sideload_file.csv
+export VK_PERFORMANCE_LAYERS_EVENT_LOG_FILE="${OUTPUT_DIR}"/events.log
+
+# Create the input file for cache sideload pipeline.
+touch "${OUTPUT_DIR}"/cache_sideload_file.csv
+
+# Test if an application can run with performance layers enabled.
+xvfb-run vkcube --c 100 && echo Run is finished successfully!
```
kuhar: nit: could you move echo to its own line and put the message in quotes? What we have right now should work, but it looks a bit weird in my editor because it treats `!` as an operator

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
...
+
+# Check if each enabled layer is writing in its corresponding log file.
```
kuhar: nit: since the application has already terminated at this point, I'd say `each enabled layer wrote/produced its log file`.

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
```
kuhar: nit: `#!/usr/bin/env bash`. The preferred bash may not be in `/bin` on all systems. For example, macOS ships a very old version of bash and the user-installed one is typically in `/usr/local/bin/`

docker/build.Dockerfile
```diff
@@ -87,12 +89,5 @@ RUN  CXX_COMPILER="g++" \
...
-RUN export LD_LIBRARY_PATH=/performance-layers/build:$LD_LIBRARY_PATH \
-    && export VK_INSTANCE_LAYERS=VK_LAYER_STADIA_pipeline_compile_time \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_runtime \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_cache_sideload \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_memory_usage \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_frame_time \
-    && export VK_LAYER_PATH=/performance-layers/layer \
-    && xvfb-run vkcube --c 100 && echo Done!
+# Enable and test the performance layers.
+RUN . /performance-layers/docker/perf_layer_env.sh /performance-layers/build /performance-layers/layer
```
kuhar: 1. Why do we need the leading `.`? I would think that we can just make the script executable (`chmod +x`) and it should work without sourcing it.
2. nit: missing newline at the end of the file
3. `perf_layer_env.sh` is not a very descriptive name. Consider something like `test_performance_layers.sh` or `check_performance_layers.sh`

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
...
+#
+# Enable the performance layers by setting the evironment variables.
```
kuhar: nit: In descriptions/documentation lines like this one we describe what an entity (function, class, module) does do, not tell it what to do. Could you `s/Enable/Enables/`, `s/Run/Runs/`, `s/Test/Tests/`?

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
...
+# Run vkcube and make sure it runs without fault when performance layers are 
+# enabled.
+# Test if each layer write to its corresponding output file. 
+
+set -e
+
+# Create a temp output directory for log files.
```
kuhar: This line is fine as-is, because the comment describes what 'we' do in this place in code, so no need to change it. Same below where I didn't comment.

kuhar (changes requested):

docker/perf_layer_env.sh
```diff
@@ -0,0 +1,60 @@
...
+# Run vkcube and make sure it runs without fault when performance layers are 
+# enabled.
+# Test if each layer write to its corresponding output file. 
+
+set -e
+
+# Create a temp output directory for log files. 
+readonly OUTPUT_DIR="$(mktemp -d)"
+readonly INSTALL_DIR="$1"
+readonly LAYER_DIR="$2"
```
kuhar: Can't we use the json files from the install dir?

docker/build.Dockerfile
miladHakimi: We either need to turn the script into an executable file or use `source`, `bash`, or `.`. I'll use the `bash` command.

kuhar: > We either need to turn the script into an executable file

This seems like the best option to me. It even already starts with a shebang suggesting it's executable

docker/build.Dockerfile
```diff
@@ -87,12 +89,8 @@ RUN  CXX_COMPILER="g++" \
...
-    && export VK_INSTANCE_LAYERS=VK_LAYER_STADIA_pipeline_compile_time \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_runtime \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_cache_sideload \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_memory_usage \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_frame_time \
-    && export VK_LAYER_PATH=/performance-layers/layer \
-    && xvfb-run vkcube --c 100 && echo Done!
+# Enable and test the performance layers.
+WORKDIR /performance-layers/docker
+RUN chmod u+x test_performance_layers.sh \
```
kuhar: Can't we change the permissions in the repo instead? You can see how the llpc CI does it: https://github.com/GPUOpen-Drivers/llpc/blob/dev/docker/llpc.Dockerfile#L37

docker/build.Dockerfile
```diff
@@ -87,12 +89,8 @@ RUN  CXX_COMPILER="g++" \
...
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_pipeline_cache_sideload \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_memory_usage \
-    && export VK_INSTANCE_LAYERS=$VK_INSTANCE_LAYERS:VK_LAYER_STADIA_frame_time \
-    && export VK_LAYER_PATH=/performance-layers/layer \
-    && xvfb-run vkcube --c 100 && echo Done!
+# Enable and test the performance layers.
+WORKDIR /performance-layers/docker
+RUN chmod u+x test_performance_layers.sh \ 
+    && ./test_performance_layers.sh \
+    /performance-layers/build
```
kuhar: not: this is not that long, no need to break this line

kuhar (approved):
Looks great. Thanks!

PR #70, opened 2022-09-26
Add checks for the content of the compile time log.

miladHakimi:
Previously, to verify the layers are working on an application, we only checked whether the log files were populated or not. This does not necessarily guarantee that the logs contain the correct data. Since we know the pattern in which the data was generated, we can check the content of the log file and see if it is as expected. There are many tools  that allow us to perform the pattern matching (e.g, LLVM FileCheck, OutputCheck, etc). We can also write a python script that reads the log file and matches it against a given pattern.
We have selected FileCheck because it is well maintained and easy to use. Moreover, we don't need to write the code that reads the input files, stores the data, etc. 
FileCheck receives two files and uses one to verify the other. Here, the `check_compile_time_log.txt` contains the patterns we expect to see in the compile_time_log.csv. The patterns can be literal (line 3) or contain a regular expression (line 4). If the patterns don't match, FileCheck will exit with a non-zero value.

### DIFF
--- a/docker/build.Dockerfile
+++ b/docker/build.Dockerfile
@@ -35,7 +35,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && export TZ=America/New_York \
     && apt-get install -yqq --no-install-recommends \
        build-essential pkg-config ninja-build \
        gcc g++ binutils-gold \
-       llvm-11 clang-11 clang-tidy-12 libclang-common-11-dev lld-11 \
+       llvm-11 clang-11 clang-tidy-12 libclang-common-11-dev lld-11 llvm-11-dev \
        python python3 python3-distutils python3-pip \
        libssl-dev libx11-dev libxcb1-dev x11proto-dri2-dev libxcb-dri3-dev \
        libxcb-dri2-0-dev lib32z1-dev libxcb-present-dev libxcb-xinerama0 libxshmfence-dev libxrandr-dev \
@@ -50,13 +50,11 @@ RUN export DEBIAN_FRONTEND=noninteractive && export TZ=America/New_York \
     && rm -rf /var/lib/apt/lists/* \
     && python3 -m pip install --no-cache-dir --upgrade pip \
     && python3 -m pip install --no-cache-dir --upgrade cmake \
-    && for tool in clang clang++ llvm-cov llvm-profdata llvm-symbolizer lld ld.lld ; do \
+    && for tool in clang clang++ llvm-cov llvm-profdata llvm-symbolizer lld ld.lld FileCheck; do \
          update-alternatives --install /usr/bin/"$tool" "$tool" /usr/bin/"$tool"-11 10 ; \
         done \
     && update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-12 10 \
     && update-alternatives --install /usr/bin/ld ld /usr/bin/ld.gold 10
-
-COPY . /performance-layers
 
 # Get all dependencies.
 WORKDIR /dependencies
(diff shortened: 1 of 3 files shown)


docker/FileCheckTests/compile_time_patterns
```diff
@@ -0,0 +1,4 @@
+; COM: Checks the pattern of compile time log file. Makes sure the header
+; COM: and the data rows' format are as expected.
```
kuhar: I think it will work even without `COM:`. There's nothing this could be confused it as this file is only passed to FileCheck.

docker/build.Dockerfile
```diff
@@ -46,7 +46,7 @@ RUN export DEBIAN_FRONTEND=noninteractive && export TZ=America/New_York \
     && wget -qO /etc/apt/sources.list.d/lunarg-vulkan-1.3.216-bionic.list https://packages.lunarg.com/vulkan/1.3.216/lunarg-vulkan-1.3.216-bionic.list \
     && apt-get update \
     && apt-get install -yqq --no-install-recommends \
-    vulkan-sdk x11-xserver-utils libvulkan-dev libvulkan1 xvfb mesa-vulkan-drivers \
+    vulkan-sdk x11-xserver-utils libvulkan-dev libvulkan1 xvfb mesa-vulkan-drivers llvm-11-dev \
```
kuhar: nit: Could you move this together with other llvm tool a few lines above?

docker/test_performance_layers.sh
```diff
@@ -55,3 +55,7 @@ echo "Run is finished successfully!"
 for file in "${output_files[@]}"; do
   check_layer_log "${file}"
 done
+
+# Check that log file's contents matches what is expected.
+FileCheck-11 FileCheckTests/compile_time_patterns --input-file \
```
kuhar: We should "install" FileCheck with `update-alternatives` so that FileCheck defaults to the currently installed version. This way when we update llvm tools we won't have to also update this script and will be able to run this script outside of CI.

docker/FileCheckTests/compile_time_patterns
```diff
@@ -0,0 +1,4 @@
+; COM: Checks the pattern of compile time log file. Makes sure the header
```
kuhar: Can we move this test outside of the docker directory? The usual convention for LLVM-based compiler projects is to put unit-tests under `unit` or `units` and FileCheck/LIT tests under `test/tests`. So I'm thinking something like `test/check_compile_time_log.txt`

kuhar (changes requested):
Great PR description, it makes what's going on very clear

docker/build.Dockerfile
miladHakimi: I did it first but I could install llvm-dev only after the apt-get update. That's why I put it here.

docker/FileCheckTests/compile_time_patterns
miladHakimi: sure.

docker/build.Dockerfile
kuhar: I've just checked and it works

kuhar: I added it just after lld-11

docker/build.Dockerfile
```diff
@@ -54,9 +54,8 @@ RUN export DEBIAN_FRONTEND=noninteractive && export TZ=America/New_York \
          update-alternatives --install /usr/bin/"$tool" "$tool" /usr/bin/"$tool"-11 10 ; \
         done \
     && update-alternatives --install /usr/bin/clang-tidy clang-tidy /usr/bin/clang-tidy-12 10 \
-    && update-alternatives --install /usr/bin/ld ld /usr/bin/ld.gold 10
-
-COPY . /performance-layers
+    && update-alternatives --install /usr/bin/ld ld /usr/bin/ld.gold 10 \
+    && update-alternatives --install /usr/bin/FileCheck FileCheck /usr/bin/FileCheck-11 10
```
kuhar: We have a for loop a few lines above that handles all llvm tools

kuhar (approved):
LGTM!

Issue #71, opened 2022-09-27
vkcube hangs with runtime layer enabled on ubuntu 22.04


kuhar:
This is under docker & lavapipe.
The other layers don't hang vkcube.

kuhar:
Running a debug build shows an assertion failure:
```
root@61641fa5d3ce:/build# timeout 5s /performance-layers/docker/test_performance_layers.sh .
Selected GPU 0: llvmpipe (LLVM 13.0.1, 256 bits), type: Cpu
WARNING: lavapipe is not a conformant vulkan implementation, testing use only.
vkcube: /performance-layers/layer/layer_data.h:237: auto performancelayers::LayerData::GetNextDeviceProcAddr(DispatchableDeviceHandleT, TFuncPtr) const [DeviceHandleT = VkCommandBuffer_T *, TFuncPtr = VkResult (*VkLayerDispatchTable_::*)(VkDevice_T *, VkQueryPool_T *, unsigned int, unsigned int, unsigned lounsigned long, unsigned int)]: Assertion `device_dispatch_iter != device_dispatch_map_.end()' failed.
```

kuhar:
Stack trace:

```
#5  0x00007fa1f260071b in __assert_fail_base (fmt=0x7fa1f27b5150 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x7fa1e8a738f6 "device_dispatch_iter != device_dispatch_map_.end()", file=0x7fa1e8a74e4b "/performance-layers/layer/layer_data.h",     line=242, function=<optimized out>) at ./assert/assert.c:92
#6  0x00007fa1f2611e96 in __GI___assert_fail (assertion=0x7fa1e8a738f6 "device_dispatch_iter != device_dispatch_map_.end()", 
    file=0x7fa1e8a74e4b "/performance-layers/layer/layer_data.h", line=242, 
[...]

PR #72, opened 2022-09-27
[CI] Simplify docker build script and upgrade OS

kuhar:
- Upgrade to ubuntu 22.04
- Upgrade llvm packages to 14
- Remove unnecessary apt and pip packages
- Build out of tree so to avoid clashes with (local) in-tree build dirs
- Delay the `COPY` command to cache dependency download & build
- Add a timeout in case runtime test hangs
- Temporarily disable the runtime layer that hangs vkcube (#71)

### DIFF
--- a/docker/build.Dockerfile
+++ b/docker/build.Dockerfile
@@ -21,7 +21,7 @@
 #                   --build-arg COMPILER=clang                   \
 #                   --build-arg GENERATOR=Ninja
 
-FROM ubuntu:20.04
+FROM ubuntu:22.04
 
 SHELL ["/bin/bash", "-c"]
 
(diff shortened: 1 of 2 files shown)


miladHakimi:
Thanks Jakub. These are really good changes. We needed to reduce the packages we were installing to make the docker faster and smaller. The Dockerfile is now more readable and easier to follow.

miladHakimi (approved):
LGTM. Let's see if @cassiebeckley has any suggestions.

kuhar:
Rebased. Could you re-approve, @miladHakimi?

docker/build.Dockerfile
```diff
@@ -72,12 +65,13 @@ RUN git clone https://github.com/KhronosGroup/Vulkan-Headers.git \
 COPY . /performance-layers
```
kuhar, line 65: This one needs to be removed

kuhar:
> Rebased. Could you re-approve, @miladHakimi?

Turns out I broke something when rebasing, I'll update this PR shortly.

kuhar:
Updated the test script to be invariant to the current working directory.

miladHakimi (approved):

PR #74, opened 2022-09-27
Add checks to the memory usage log file content.

miladHakimi:
`check_memory_usage_log.txt` checks the header and the data rows' format. It also checks the data rows' count using `CHECK-COUNT-<num>` and `CHECK-NOT` to make sure the row count is exactly as expected.


kuhar (approved):
Nice, I didn't know about `-COUNT` checks

PR #76, opened 2022-09-28
Add checks to the frame time log file content.

miladHakimi:
### `check_frame_time_log.txt` checks the header and the data rows' format. It also checks the data rows' count using `CHECK-COUNT-<num>` and `CHECK-NOT` to make sure the number is exactly as expected.


kuhar (approved):
LGTM

Issue #75, opened 2022-09-28
Introduce abstractions for loggable events


kuhar:
The existing event logging is rather ad-hoc. Examples:
```c++
  FrameTimeLayerData(char* log_filename, uint64_t exit_frame_num_or_invalid,
                     const char* benchmark_watch_filename,
                     const char* benchmark_start_string)
      : LayerData(log_filename, "Frame Time (ns),Benchmark State"),
        exit_frame_num_or_invalid_(exit_frame_num_or_invalid),
        benchmark_start_pattern_(StrOrEmpty(benchmark_start_string)) {
    LogEventOnly("frame_time_layer_init");
```

```c++
  layer_data->LogTimeDelta("frame_present",
                           layer_data->HasBenchmarkStarted() ? "1" : "0");
```

```c++
    int64_t logged_delta = ToInt64Nanoseconds(now - last_log_time_);
    LogLine(event_type, CsvCat(logged_delta, extra_content), now);
```

```c++
    layer_data->LogEventOnly("frame_time_layer_exit",
                             absl::StrCat("terminated,frame:", frames_elapsed));
```

This form of logging is unstructured and does not allow for emitting logs in different output formats. All the event data is converted to strings and concatenated before we can decide which logging format to use.

This task is to come up with a more structured abstraction for logged events that will allow for multiple output formats (e.g., csv and json).

The generic interface should provide the following functions:
1.  `getEventName()`
2.  `getLogLevel()` -- decides the 'verbosity level', e.g., whether to log to CSV or to the common log file 
3.  `getNumAttributes` -- returns the number of data entries (or 'columns' / tuple elements / etc)
5.  `getAttributes()`

We want to support a few data entry types:
1. String
2. Int64  -- e.g., count of something
3. Int64Array -- e.g., array of hashes
4. Timepoint / timestamp -- when something happened
5. Duration -- how long something took

All of this combines as follows:
* Event is a names set of **attributes**
* Attribute is a key-value pair where the attribute name is the key and value can be of one of the supported types.

A concrete layer-specific event definition could look something like this:
```c++
class ShaderModuleCreateEvent : public LoggableEvent {
public:
   // (Constructor and other functions...)
   
  std::vector<const Attribute*> getAttributes() const {
    return {&shader_module_hashes_, &creation_time_};
  }
  
private:
  Int64ArrayAttr shader_module_hashes_("shader_hashes");
  DurationAttr creation_time_("creation_time");
};

[...]

PR #79, opened 2022-10-04
Implement Event class for creating logs. 

miladHakimi:
Currently, the logging mechanism is unstructured. Various log functions with different interfaces are used to generate the logs. All these functions receive a series of inputs, combine them in a string with csv format, and write them in the standard output or a file.
`The`Event` class unifies the interface for creating logs. It provides these two functionalities:
1) Using the same interface for creating all the logs to replace the current ad-hoc event logging functions.
2) Allowing log generation for different formats other than csv.

Each `Event` has a name and a set of attributes. Attributes contain the value and its name. The name is a string telling what the attribute is (e.g, duration, timestamp, etc).

The current log functions use only a limited set of types; string, int64, and vector. Hence, right now, `Attribute` has three implementations covering the mentioned types.

This PR also contains the implementation for Event, an event-to-csv generator, and simple unit tests to check if they work as expected.

Issue: #75

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -133,6 +133,7 @@ install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/layer
 
 enable_testing()
 add_executable(layer_support_tests
+    units/event_log_tests.cc
     units/input_buffer_tests.cc
     units/log_scanner_tests.cc
 )
--- a/units/event_log_tests.cc
+++ b/units/event_log_tests.cc
@@ -1,0 +1,275 @@
+#include <cstddef>
+#include <cstdint>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include "gtest/gtest.h"
+
+namespace {
+// Set of supported value types for an attribute.
+enum ValueType { String, Int64, VectorInt64 };
+// Specifies the importance of an Event. Events are logged based on their level
+// of importance. E.g., compile_time.csv only contains the important compile
+// time events while event.log contains all the compile time events.
+enum LogLevel { Low, Medium, High };
+
+// Attribute class
+//
+// An event consists of a set of attributes. Each attribute has a name that
+// indicates what the attribute is (e.g., timestamp, hash, etc) and a value.
+// `Attribute` is the base class for an attribute. For each `ValueType`, there
+// should be an implmentation of `Attribute`. Each implementation must expose an
+// `id_` to indicate which ValueType it is implementing.
+class Attribute {
+ public:
+  // Checks if the `Attribute` is an instance of the given class. Used for
+  // safely casting an `Attribute` to a proper derived class.
+  template <class T>
+  bool isa() {
+    return value_type_ == T::id_;
+  }
+
+  // Safely casts the Attribute to the given class (T). Valid typenames should
+  // be derived from Attribute and have an ValueType `id_` field.
+  template <class T>
+  const T *cast() {
+    return isa<T>() ? static_cast<T *>(this) : nullptr;
+  };
+
+  const char *GetName() const { return name_; }
+  ValueType GetValueType() const { return value_type_; }
+
+ protected:
+  Attribute(const char *name, ValueType value_type)
+      : name_(name), value_type_(value_type) {}
+
+ private:
+  const char *name_;
+  ValueType value_type_;
+};
+
+// Implements the Attribute base class for the `ValueType`s.
+// Valid typenames: {std::string, int64_t, std::vector<int64_t>}
+// Example:
+//    `AttributeImpl<int64_t, ValueType::Int64> timestamp("timestamp", 1234);`
+template <typename T, ValueType VT>
+class AttributeImpl : public Attribute {
+ public:
+  static constexpr ValueType id_ = VT;
+
+  AttributeImpl(const char *name, const T &value)
+      : Attribute{name, VT}, value_(value) {}
+
+  const T &GetValue() const { return value_; }
+
+ protected:
+  T &GetValue() { return value_; }
+
+ private:
+  T value_;
+};
+
+using VectorInt64Attr =
+    AttributeImpl<std::vector<int64_t>, ValueType::VectorInt64>;
+using StringAttr = AttributeImpl<std::string, ValueType::String>;
+using Int64Attr = AttributeImpl<int64_t, ValueType::Int64>;
+
+// Event represents the base struct for a loggable event. It contains the
+// event's name and the level of importance. The derived structs must define and
+// initialize their own set of attributes.
+class Event {
+ public:
+  Event(const char *name, LogLevel log_level)
+      : name_(name), log_level_(log_level) {}
+
+  // Each implementation of an `Event` contains attribute(s) and
+  // overrides this function to return them.
+  virtual const std::vector<Attribute *> &GetAttributes() = 0;
+  virtual size_t GetNumAttributes() const = 0;
+
+  const char *GetEventName() const { return name_; }
+  LogLevel GetLogLevel() const { return log_level_; }
+
+ private:
+  const char *name_;
+  LogLevel log_level_;
+};
+
+// An `Event` for the CreateShaderModule function.
+// Example:
+// ```c++
+//    const int64_t create_time_ns = create_end - create_start;
+//    Int64Attr timestamp("timestamp", 1234);
+//    Int64Attr shader_hash("hash", res.shader_hash);
+//    std::vector<Attribute *> attributes = {&shader_hash, &timestamp};
+//    CreateShaderModuleEvent("create_shader_module_ns", attributes,
+//    LogLevel::High);
+// ```
+struct CreateShaderModuleEvent : Event {
+ public:
+  CreateShaderModuleEvent(const char *name, int64_t timestamp,
+                          int64_t hash_value, int64_t duration,
+                          LogLevel log_level)
+      : Event(name, log_level),
+        timestamp_{"timestamp", timestamp},
+        hash_value_{"hash", hash_value},
+        duration_{"duration", duration},
+        attributes_{&timestamp_, &hash_value_, &duration_} {}
+
+  const std::vector<Attribute *> &GetAttributes() override {
+    return attributes_;
+  }
+  size_t GetNumAttributes() const override { return attributes_.size(); }
+
+ private:
+  Int64Attr timestamp_;
+  Int64Attr hash_value_;
+  Int64Attr duration_;
+  std::vector<Attribute *> attributes_;
+};
+
+struct CreateGraphicsPipelinesEvent : Event {
+ public:
+  CreateGraphicsPipelinesEvent(const char *name, int64_t timestamp,
+                               VectorInt64Attr &hash_values, int64_t duration,
+                               LogLevel log_level)
+      : Event(name, log_level),
+        timestamp_{"timestamp", timestamp},
+        hash_values_(hash_values),
+        duration_{"duration", duration},
+        attributes_{&timestamp_, &hash_values_, &duration_} {}
+
+  const std::vector<Attribute *> &GetAttributes() override {
+    return attributes_;
+  }
+  size_t GetNumAttributes() const override { return attributes_.size(); }
+
+ private:
+  Int64Attr timestamp_;
+  VectorInt64Attr hash_values_;
+  Int64Attr duration_;
+  std::vector<Attribute *> attributes_;
+};
+
+std::string ValueToCSVString(const std::string &value) { return value; }
+
+std::string ValueToCSVString(const int64_t value) {
+  return std::to_string(value);
+}
+
+std::string ValueToCSVString(const std::vector<int64_t> &values) {
+  std::stringstream csv_string;
+  csv_string << "\"[";
+  size_t e = values.size();
+  for (size_t i = 0; i != e; ++i) {
+    const char *delimiter = i < e - 1 ? "," : "";
+    csv_string << values[i] << delimiter;
+  }
+  csv_string << "]\"";
+  return csv_string.str();
+}
+
+// Takes an `Event` instance as an input and generates a csv string containing
+// `event`'s name and attribute values.
+// TODO(miladhakimi): Differentiate addresses and other integers. Addresses
+// should be displayed in hex.
+std::string EventToCSVString(Event &event) {
+  const std::vector<Attribute *> &attributes = event.GetAttributes();
+
+  std::stringstream csv_str;
+  csv_str << event.GetEventName();
+  csv_str << ",";
+  for (size_t i = 0, e = attributes.size(); i != e; ++i) {
+    switch (attributes[i]->GetValueType()) {
+      case ValueType::Int64: {
+        csv_str << ValueToCSVString(
+            attributes[i]->cast<Int64Attr>()->GetValue());
+        break;
+      }
+      case ValueType::String: {
+        csv_str << ValueToCSVString(
+            attributes[i]->cast<StringAttr>()->GetValue());
+        break;
+      }
+      case ValueType::VectorInt64: {
+        csv_str << ValueToCSVString(
+            attributes[i]->cast<VectorInt64Attr>()->GetValue());
+        break;
+      }
+    }
+    if (i + 1 != e) csv_str << ",";
+  }
+  return csv_str.str();
+}
+
+TEST(Event, AttributeCreation) {
+  const int64_t timestamp_val = 1601314732230797664;
+  const int64_t hash_val1 = 0x67d6fd0aaa78a6d8;
+  const int64_t hash_val2 = 0x67d390249c2f20ce;
+  const Int64Attr timestamp("timestamp", timestamp_val);
+  const StringAttr state("state", "1");
+  const VectorInt64Attr pipeline("pipeline", {hash_val1, hash_val2});
+  EXPECT_EQ(timestamp.GetName(), "timestamp");
+  EXPECT_EQ(timestamp.GetValue(), timestamp_val);
+  EXPECT_EQ(state.GetName(), "state");
+  EXPECT_EQ(state.GetValue(), "1");
+  EXPECT_EQ(pipeline.GetName(), "pipeline");
+  ASSERT_EQ(pipeline.GetValue().size(), 2);
+  EXPECT_EQ(pipeline.GetValue()[0], hash_val1);
+  EXPECT_EQ(pipeline.GetValue()[1], hash_val2);
+}
+
+TEST(Event, CreateShaderModuleEventCreation) {
+  const int64_t timestamp_val = 1601314732230797664;
+  const int64_t hash_val1 = 0x67d6fd0aaa78a6d8;
+  const int64_t duration = 926318;
+  CreateShaderModuleEvent compile_event("compile_time", timestamp_val,
+                                        hash_val1, duration, LogLevel::Low);
+  EXPECT_EQ(compile_event.GetNumAttributes(), 3);
+}
+
+TEST(Event, ShaderModuleEventCSVStringGeneration) {
+  const int64_t timestamp_val = 1601314732230797664;
+  const int64_t hash_val1 = 0x67d6fd0aaa78a6d8;
+  const int64_t duration = 926318;
+  std::stringstream expected_str;
+  expected_str << "compile_time," << timestamp_val << "," << hash_val1 << ","
+               << duration;
+  CreateShaderModuleEvent compile_event("compile_time", timestamp_val,
+                                        hash_val1, duration, LogLevel::Low);
+  ASSERT_EQ(compile_event.GetNumAttributes(), 3);
+
+  EXPECT_EQ(EventToCSVString(compile_event), expected_str.str());
+}
+
+TEST(Event, GraphicsPipelinesEventCSVStringGeneration) {
+  const int64_t timestamp_val = 1601314732230797664;
+  const int64_t hash_val1 = 0x67d6fd0aaa78a6d8;
+  const int64_t hash_val2 = 0x67d390249c2f20ce;
+  const int64_t duration = 926318;
+
+  VectorInt64Attr hashes("hashes", {hash_val1, hash_val2});
+  CreateGraphicsPipelinesEvent pipeline_event("create_graphics_pipeline",
+                                              timestamp_val, hashes, duration,
+                                              LogLevel::Low);
+  EXPECT_EQ(pipeline_event.GetNumAttributes(), 3);
+}
+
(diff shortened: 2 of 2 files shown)


units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
+#include <array>
```
kuhar: Could you remove this and any other unused includes?

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+#include <cstdio>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "gtest/gtest.h"
+
```
kuhar, line 8: Could you put all the code below in an anonymous namespace? This way we won't have to worry about any name clashes and [ODR violations](https://en.wikipedia.org/wiki/One_Definition_Rule).

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+// TODO(miladhakimi): Need better naming.
+enum LogLevel { CommonFile, PrivateFile };
+enum ValueType { String, Int64, Vector };
```
kuhar: Can `Vector` be used to compose other `ValueType`s?  If Vector can only hold one type, should this be `VectorInt64` instead?

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+#include <cstdint>
+#include <cstdio>
+#include <iostream>
+#include <string>
+#include <vector>
+
+#include "gtest/gtest.h"
+
+// TODO(miladhakimi): Need better naming.
+enum LogLevel { CommonFile, PrivateFile };
```
kuhar: Could you add a short comment for this and other types/functions you plan to expose in the final public interface?

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+
+std::string ValueToString(const std::string &value) { return value; }
+
+std::string ValueToString(const int64_t &value) {
+  return std::to_string(value);
+}
+
+// TODO(miladhakimi): This can be different for other formats.
+template <typename T>
+std::string ValueToString(const T &values) {
```
kuhar: I think it would be better to introduce an overload for `vector<int64_t>` only. This will simplify the code and won't make others think that more types are supported

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+
+#include "gtest/gtest.h"
+
+// TODO(miladhakimi): Need better naming.
+enum LogLevel { CommonFile, PrivateFile };
+enum ValueType { String, Int64, Vector };
+
+std::string ValueToString(const std::string &value) { return value; }
+
+std::string ValueToString(const int64_t &value) {
```
kuhar: Ints are cheap to copy, pass it by value

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+  return std::to_string(value);
+}
+
+// TODO(miladhakimi): This can be different for other formats.
+template <typename T>
+std::string ValueToString(const T &values) {
+  std::string tmp = "\"[";
```
kuhar: `std::stringstream` should be more efficient (in terms of the number of allocations) for incrementally building long strings.

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+  return tmp;
+}
+
+struct Attribute {
+  const char *name;
+  ValueType value_type;
+
+  template <typename T>
+  T *cast() {
+    return static_cast<T *>(this);
```
kuhar: This looks unsafe, we don't check if the cast is possible/legal. We should either check that the value type matches the requested type `T` or drop this function entirely and perform `static_cast`s directly

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+  T &GetValue() { return value; }
+};
+
+using VectorAttr = AttributeImpl<std::vector<int64_t>, ValueType::Vector>;
+using StringAttr = AttributeImpl<std::string, ValueType::String>;
+using Int64Attr = AttributeImpl<int64_t, ValueType::Int64>;
+
+struct EventLog {
+  const char *name;
+  std::vector<Attribute *> attributes;
```
kuhar: I think this class matches the 'LogabbleEvent` class in the original issue (https://github.com/googlestadia/performance-layers/issues/69), but the name suggests it's the log of events. This seems misleading to me.

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+  LogLevel log_level;
+
+  EventLog(const char *_name, const std::vector<Attribute *> &_attributes,
+           LogLevel _log_level)
+      : name(_name), attributes(_attributes), log_level(_log_level) {}
+
+  const char *GetEventName() { return name; }
+  const std::vector<Attribute *> &GetAttributes() { return attributes; }
```
kuhar: What do you think about making this function `virtual` and letting the derived classes override it? The reason is that when some class derives from `EventLog` it must first run an `EventLog`'s constructor. If attributes are members in the derived class, you may not get the change to properly initialize them before you have to call the base constructor with attribute pointers.

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+std::string EventToCSVString(EventLog &event) {
+  const std::vector<Attribute *> &attributes = event.GetAttributes();
+
+  std::string csv_str = event.GetEventName();
+  csv_str += ",";
+  size_t e;
+  for (size_t i = 0, e = attributes.size(); i != e; ++i) {
+    switch (attributes[i]->value_type) {
+      case ValueType::Int64: {
+        csv_str += ValueToString(attributes[i]->cast<Int64Attr>()->GetValue());
```
kuhar: This `ValueToString` implementation/format seems specific to CSV, especially when we consider array printing. What do you think about calling it `ValueToCSVString` instead?

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+      }
+      case ValueType::String: {
+        csv_str += ValueToString(attributes[i]->cast<StringAttr>()->GetValue());
+        break;
+      }
+      case ValueType::Vector: {
+        csv_str += ValueToString(attributes[i]->cast<VectorAttr>()->GetValue());
+        break;
+      }
+      default:
```
kuhar: Some compilers warn about a default switch cases when the other cases are already exhaustive. I think we should remove this default case and instead rely on compiler warnings.

units/event_log_tests.cc
```diff
@@ -0,0 +1,132 @@
...
+    if (i + 1 != e) csv_str += ",";
+  }
+  return csv_str;
+}
+
+TEST(EventLog, AttributeCreation) {
+  Int64Attr compile_timestamp("timestamp", 1601314733266018318);
+  StringAttr state("state", "1");
+  VectorAttr pipeline("pipeline", {0x67d6fd0aaa78a6d8, 0x67d390249c2f20ce});
+  SUCCEED();
```
kuhar: This doesn't seem to check anything. Maybe we could check that the number of attributes is as expected and that the attribute pointers match the local variables?

kuhar (changes requested):
> The current log functions use only a limited set of types; string, int64, and vector. Hence, right now, Attribute has three implementations covering the mentioned types.

> Fixes: https://github.com/googlestadia/performance-layers/issues/75

This PR doesn't fully fix the issue, we should use `Issue: <link>` instead

units/event_log_tests.cc
miladHakimi: Makes sense.

units/event_log_tests.cc
miladHakimi: Yeah. True. I put a comment on top to fix it. The name is general but the functionality is specific to csv string.

CMakeLists.txt
```diff
@@ -135,6 +135,7 @@ enable_testing()
 add_executable(layer_support_tests
     units/input_buffer_tests.cc
     units/log_scanner_tests.cc
+    units/event_log_tests.cc
```
kuhar: Could you put this unit test file first? It's supposed to be in the alphabetical order.

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
+#include <cassert>
+#include <cstddef>
+#include <cstdint>
+#include <cstdio>
+#include <iostream>
+#include <ostream>
```
kuhar: I don't see these three includes used, can we drop them?

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+std::string ValueToCSVString(const int64_t value) {
+  return std::to_string(value);
+}
+
+std::string ValueToCSVString(const std::vector<int64_t> &values) {
+  std::string tmp = "\"[";
+  using std::begin;
+  using std::end;
+  auto e = end(values);
+  for (auto it = begin(values); it != e; ++it) {
```
kuhar: This loop no longer needs to be generic, we should change this to a plain for loop over all indices. Something like:
```c++
for (size_t i  = 0, e = values.end(); i != e; ++i)
```

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+  tmp += "]\"";
+  return tmp;
+}
+
+struct Attribute {
+  const char *name;
+  AttributeType value_type;
+
+  template <typename T>
+  bool isa() {
```
kuhar: Could you add a comment of explanation for this function?

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+  for (auto it = begin(values); it != e; ++it) {
+    const char *delimiter = std::next(it) != values.end() ? "," : "";
+    tmp += std::to_string(*it) + delimiter;
+  }
+  tmp += "]\"";
+  return tmp;
+}
+
+struct Attribute {
+  const char *name;
```
kuhar: I think it would be better for `Attribute` to be a `class` with clearly defined public and private interface. For example, we probably don't want all users of `Attribute` to be able to change the name, which we can enforce by making the field `private` and exposing a getter `getName() const;`.

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+  AttributeType value_type;
+
+  template <typename T>
+  bool isa() {
+    return value_type == T::id;
```
kuhar: Because this code requires all derived classes to expose `id`, I think we should document this in a comment above `Attribute`

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+  auto e = end(values);
+  for (auto it = begin(values); it != e; ++it) {
+    const char *delimiter = std::next(it) != values.end() ? "," : "";
+    tmp += std::to_string(*it) + delimiter;
+  }
+  tmp += "]\"";
+  return tmp;
+}
+
+struct Attribute {
```
kuhar: Documentation is missing

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+// Valid typenames: {std::string, int64_t, std::vector<int64_t>}
+// Example:
+//    AttributeImpl<int64_t, AttributeType::Int64> timestamp("timestamp", 1234);
+template <typename T, AttributeType VT>
+struct AttributeImpl : Attribute {
+  T value;
+  static constexpr AttributeType id = VT;
+
+  AttributeImpl(const char *_name, T _value)
+      : Attribute{_name, VT}, value(_value) {}
```
kuhar: nit: the names don't conform to the style guide: https://google.github.io/styleguide/cppguide.html#Naming
This should be `(const char *name, T value)`

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+// Implements the Attribute base class for each specific `AttributeType`.
+// Valid typenames: {std::string, int64_t, std::vector<int64_t>}
+// Example:
+//    AttributeImpl<int64_t, AttributeType::Int64> timestamp("timestamp", 1234);
+template <typename T, AttributeType VT>
+struct AttributeImpl : Attribute {
+  T value;
+  static constexpr AttributeType id = VT;
+
+  AttributeImpl(const char *_name, T _value)
```
kuhar: Because this code is generic and will handle containers like `vector` or `string`, we should avoid needless copying.   We should take the value by a const reference.

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+  };
+};
+
+// Implements the Attribute base class for each specific `AttributeType`.
+// Valid typenames: {std::string, int64_t, std::vector<int64_t>}
+// Example:
+//    AttributeImpl<int64_t, AttributeType::Int64> timestamp("timestamp", 1234);
+template <typename T, AttributeType VT>
+struct AttributeImpl : Attribute {
+  T value;
```
kuhar: Similarly, could you make this a class and disallow modification to the underlying value? I would introduce a public const getter `const T& GetValue() const` and a non-const protected for derived classes `T& GetValue()`.

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+  template <typename T>
+  T *cast() {
+    return isa<T>() ? static_cast<T *>(this) : nullptr;
+  };
+};
+
+// Implements the Attribute base class for each specific `AttributeType`.
+// Valid typenames: {std::string, int64_t, std::vector<int64_t>}
+// Example:
+//    AttributeImpl<int64_t, AttributeType::Int64> timestamp("timestamp", 1234);
```
kuhar: Thanks, these examples look very useful. Could you surround them with code block, similar to `AttributeType` above?

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+    AttributeImpl<std::vector<int64_t>, AttributeType::VectorInt64>;
+using StringAttr = AttributeImpl<std::string, AttributeType::String>;
+using Int64Attr = AttributeImpl<int64_t, AttributeType::Int64>;
+
+// Event represents the base struct for a loggable event. It contains the
+// event's name and the level of importance. The derived structs must define and
+// initialize their own set of attributes.
+struct Event {
+  const char *name;
+  LogLevel log_level;
```
kuhar: Same here: class + deliberate access specifiers

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+};
+
+// An event for the CreateShaderModule function.
+// Example:
+//    const int64_t create_time_ns = create_end - create_start;
+//    Int64Attr timestamp("timestamp", 1234);
+//    Int64Attr shader_hash("hash", res.shader_hash);
+//    std::vector<Attribute *> attributes = {&shader_hash, &timestamp};
+//    CreateShaderModuleEvent("create_shader_module_ns", attributes,
+//    LogLevel::High);
```
kuhar, line 107: Could you surround this example with a code block?
````
```c++
your code
```
````

File: units/event_log_tests.cc [hunk 27]
```diff
@@ -0,0 +1,183 @@
...
+struct CreateShaderModuleEvent : Event {
+  std::vector<Attribute *> attributes;
+
+  CreateShaderModuleEvent(const char *_name,
+                          std::vector<Attribute *> &_attributes,
```
kuhar: I think it would be better to take make this event own all attributes. This would imply having a field per each attribute and setting them in the constructor.

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+  const std::vector<Attribute *> &GetAttributes() override {
+    return attributes;
+  }
+  size_t GetNumAttributes() override { return attributes.size(); }
+};
+
+// Takes an `Event` instance as an input and generates a csv string containing
+// `event`'s name and attribute values.
+std::string EventToCSVString(Event &event) {
+  const std::vector<Attribute *> &attributes = event.GetAttributes();
```
kuhar: Could you move all the `ValueToCSVString` just before this function? Keeping related code together will make it easier to understand.

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+
+  std::string csv_str = event.GetEventName();
+  csv_str += ",";
+  size_t e;
+  for (size_t i = 0, e = attributes.size(); i != e; ++i) {
+    switch (attributes[i]->value_type) {
+      case AttributeType::Int64: {
+        csv_str +=
+            ValueToCSVString(attributes[i]->cast<Int64Attr>()->GetValue());
```
kuhar: Use `std::stringstream` to build this string

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+        break;
+      }
+    }
+    if (i + 1 != e) csv_str += ",";
+  }
+  return csv_str;
+}
+
+TEST(Event, AttributeCreation) {
+  Int64Attr timestamp("timestamp", 1601314733266018318);
```
kuhar: In this basic test we could check that the value and name can be retrieved and that they match what you passed to create each of the attributes

units/event_log_tests.cc
```diff
@@ -0,0 +1,183 @@
...
+
+TEST(Event, EventCreation) {
+  Int64Attr timestamp("timestamp", 1601314733266018318);
+  StringAttr state("state", "1");
+  VectorInt64Attr pipeline("pipeline",
+                           {0x67d6fd0aaa78a6d8, 0x67d390249c2f20ce});
+  std::vector<Attribute *> compile_attrs = {&timestamp, &state, &pipeline};
+  CreateShaderModuleEvent compile_event("compile_time", compile_attrs,
+                                        LogLevel::Low);
+  EXPECT_EQ(compile_event.GetNumAttributes(), 3);
```
kuhar, line 229: Use `ASSERT_EQ` instead. The difference is that `ASSERT_*` checks abort execution of the current test on failure, while `CHECK_*` continue executing. When the size is not as expected, we shouldn't try to access vector elements that don't exist.

units/event_log_tests.cc
miladHakimi: Yeah. I wanted to do this in the next steps.

units/event_log_tests.cc
```diff
@@ -0,0 +1,213 @@
...
+
+  // Safely casts the Attribute to the given class (T). Valid typenames should
+  // be derived from Attribute and have an ValueType `id_` field.
+  template <class T>
+  const T *cast() {
+    return isa<T>() ? static_cast<T *>(this) : nullptr;
+  };
+
+  const char *GetName() const { return name_; }
+  ValueType GetValueType() const { return value_type_; }
```
kuhar, line 41: Use `const ValueType& GetValueType() const` to avoid unnecessary copies

units/event_log_tests.cc
```diff
@@ -0,0 +1,213 @@
...
+  return std::to_string(value);
+}
+
+std::string ValueToCSVString(const std::vector<int64_t> &values) {
+  std::stringstream csv_string;
+  csv_string << "\"[";
+  size_t e = values.size();
+  for (size_t i = 0; i != e; ++i) {
+    const char *delimiter = i < e - 1 ? "," : "";
+    csv_string << std::to_string(values[i]) + delimiter;
```
kuhar: No need to `to_string`, we can just append directly:
```c++
csv_string << values[i] << delimiter;
```

units/event_log_tests.cc
```diff
@@ -0,0 +1,213 @@
...
+
+// Takes an `Event` instance as an input and generates a csv string containing
+// `event`'s name and attribute values.
+std::string EventToCSVString(Event &event) {
+  const std::vector<Attribute *> &attributes = event.GetAttributes();
+
+  std::stringstream csv_str;
+  csv_str << event.GetEventName();
+  csv_str << ",";
+  size_t e;
```
kuhar: No need to define `e` here, the `for` loop below already defines it

units/event_log_tests.cc
```diff
@@ -0,0 +1,213 @@
...
+  return csv_str.str();
+}
+
+TEST(Event, AttributeCreation) {
+  const Int64Attr timestamp("timestamp", 1601314733266018318);
+  const StringAttr state("state", "1");
+  const VectorInt64Attr pipeline("pipeline",
+                                 {0x67d6fd0aaa78a6d8, 0x67d390249c2f20ce});
+  EXPECT_EQ(timestamp.GetName(), "timestamp");
+  EXPECT_EQ(timestamp.GetValue(), 1601314733266018318);
```
kuhar: It'd be better to put those long constants in local variables, so that it's immediately obvious when things should be equal, e.g.:
```c++
const int64_t timestamp_val = 160...;
const Int64Attr timestamp("timestamp", timestamp_val);
...
CHEKC_EQ(timestemp.GetValue(), timestamp_val);
```

units/event_log_tests.cc
```diff
@@ -0,0 +1,213 @@
...
+  ASSERT_EQ(pipeline.GetValue().size(), 2);
+  EXPECT_EQ(pipeline.GetValue()[0], 0x67d6fd0aaa78a6d8);
+  EXPECT_EQ(pipeline.GetValue()[1], 0x67d390249c2f20ce);
+}
+
+TEST(Event, CreateShaderModuleEventCreation) {
+  CreateShaderModuleEvent compile_event("compile_time", 1601314733266018318,
+                                        0x67d6fd0aaa78a6d8, 926318,
+                                        LogLevel::Low);
+  ASSERT_EQ(compile_event.GetNumAttributes(), 3);
```
kuhar: Here `EXPECT_EQ` is fine because there is no other check that would otherwise fail.

kuhar (commented):
The PR looks very nice now, just a few minor things

units/event_log_tests.cc
miladHakimi: Since the ValueType's size was 4 bytes, I assumed the copy would be cheap. Sure, I'll change it.

kuhar: Ohhh, I got it wrong, I thought that this returns the underlying value. Please disregard then.

units/event_log_tests.cc
```diff
@@ -0,0 +1,275 @@
...
+
+  // Safely casts the Attribute to the given class (T). Valid typenames should
+  // be derived from Attribute and have an ValueType `id_` field.
+  template <class T>
+  const T *cast() {
+    return isa<T>() ? static_cast<T *>(this) : nullptr;
+  };
+
+  const char *GetName() const { return name_; }
+  const ValueType &GetValueType() const { return value_type_; }
```
kuhar: Could you undo this change? I thought that this returned the underlying value, not the enum. Sorry for the confusion!

kuhar (approved):
Awesome, LGTM

PR #80, opened 2022-10-06
Refactor the namespaces in input_buffer_tests.cc.

miladHakimi:
This file contains nested anonymous namespaces. The PR removes the inner namespace because it was unnecessary.


kuhar (approved):
Thanks!

Issue #81, opened 2022-10-08
Introduce event logging abstraction.


miladHakimi:
An EventLogger takes a stream of `Event` as input and produces a log based on the desired format (CSV, Chrome trace event, etc).
 
```mermaid
flowchart LR
    EventCreator --Event --> EventLogger --log--> OutputLogger
```
The EventLogger would be an abstract class with some public methods: `AddEvent()`,  `StartLog()`, and `FinalizeLog()`. It also requires a flush mechanism. Either a public `Flush()` method triggered from outside, or an implicitly triggered `Flush()` invoked when the number of current events reaches a certain threshold. 
Each format may require prefixes and suffixes around the final log. For instance, CSV requires adding a header to the first line of the file. `StartLog()` and `FinalizeLog()` should be implemented according to the specifications of each format. 

To generate a log, these steps are required:
1. Converting each `Event` into an entry: 
Each `Event` is an entry in the final log output. The entry format is defined by the EventLogger implementation. For instance: The entry for `Event(name: "compile_time_ns", timestamp: 1234, hash:0xabc, duration:123)` would look like this in the CSV format:
```
compile_time,1234,0xabc,123
```
and this in the Chrome event trace format:
```
{name:"compile_time_ns", timestamp: 1234, hash: 0xabc, duration: 123}
```
2. Combining entries:
This step, generates the log's body by combining the entries. In CSV, the entries are separated by a new line and in Chrome event trace, they're comma separated.
The pseudo code for the EventLogger would be like this:

``` c++
class EventLogger {
 public:
  void InsertEvent(Event* e) = 0;
  // Flush may receive an input (an instance of LogOutput), pass the string generated from concatenating the events to it, and 
  // delete the current events.
  void Flush(...) = 0;
  void StartLog() = 0;
  void FinishLog() = 0;
 protected:
  std::vector<Event*> events_;
};

class EventLoggerImpl: public EventLogger {
 public:
  void Flush(...) override {
    // Convert Events to entries.
    ...
    // Combine entries.
    ...
    // Delete current events
    ...
    // Use LogOutput to print the log
  }
}
```

kuhar:
Thanks for writing this up!

A few comments and questions:

> An EventLogger takes Event(s) as input and produces a log based on the desired format (CSV, Chrome trace event, etc).

We could say that the input is a stream or sequence of Events.

[...]

PR #84, opened 2022-10-18
Implement Logger abstraction for logging Events.

miladHakimi:
`EventLogger` takes `Event`s as input and produces logs based on a desired format. Each output format (such as CSV, ChromeTraceEvent, etc) implements the `EventLogger` and overrides the virtual methods based on its specifications.

`FilterLogger` receives an `EventLogger` and a `LogLevel` and filters the incoming events to the logger based on the given `LogLevel`.

`BroadcastLogger` is a logger that keeps multiple loggers. When a function is invoked, it invokes the function on all of its loggers. It allows us to register multiple loggers for a specific event. For instance, compile time logs can be in the compile_time private file and the event.log file. Hence, we register both loggers for the compile time events.

Issue: #81

### DIFF
--- a/units/event_log_tests.cc
+++ b/units/event_log_tests.cc
@@ -4,7 +4,11 @@
 #include <string>
 #include <vector>
 
+#include "gmock/gmock.h"
 #include "gtest/gtest.h"
+
+using ::testing::ElementsAre;
+using ::testing::IsEmpty;
 
 namespace {
 // Set of supported value types for an attribute.
@@ -174,7 +178,7 @@ std::string ValueToCSVString(const std::vector<int64_t> &values) {
 
 // Takes an `Event` instance as an input and generates a csv string containing
 // `event`'s name and attribute values.
-// TODO(miladhakimi): Differentiate addresses and other integers. Addresses
+// TODO(miladhakimi): Differentiate hashes and other integers. Hashes
 // should be displayed in hex.
 std::string EventToCSVString(Event &event) {
   const std::vector<Attribute *> &attributes = event.GetAttributes();
@@ -204,6 +208,119 @@ std::string EventToCSVString(Event &event) {
   }
   return csv_str.str();
 }
+
+// EventLogger base class
+//
+// EventLogger provides an abstraction for the concrete loggers that log the
+// events in various formats (CSV, Chrome Trace Event, etc). The `EventLogger`'s
+// methods can be called from multiple threads at the same time. Hence, they are
+// expected to be internally synchronized.
+class EventLogger {
+ public:
+  virtual ~EventLogger() = default;
+
+  // Converts an `Event` into a string and writes it to the output.
+  virtual void AddEvent(Event *event) = 0;
+
+  // This method is called exactly once before any events are added to denote
+  // that the log has started. For instance, in case of a CSV log, the CSV
+  // header is written when this function is called.
+  virtual void StartLog() = 0;
+
+  // This method is called exactly once to denote that the log is finished.
+  virtual void EndLog() = 0;
+
+  // Makes sure all the logs are written to the output.
+  virtual void Flush() = 0;
+};
+
+// FilterLogger is a wrapper around the `EventLogger` that filters the input
+// events based on a given log level.
+// Example:
+// ```C++
+// CSVLogger logger = ...;
+// FilterLogger filter(&logger, LogLevel::High);
+// ```
+class FilterLogger : public EventLogger {
+ public:
+  FilterLogger(EventLogger *logger, LogLevel log_level)
+      : logger_(logger), log_level_(log_level){};
+
+  void AddEvent(Event *event) override {
+    if (event->GetLogLevel() >= log_level_) logger_->AddEvent(event);
+  }
+
+  void StartLog() override { logger_->StartLog(); }
+
+  void EndLog() override { logger_->EndLog(); }
+
+  void Flush() override { logger_->Flush(); }
+
+ private:
+  EventLogger *logger_ = nullptr;
+  LogLevel log_level_ = LogLevel::Low;
+};
+
+// A subclass of `EventLogger` that forwards all events, start/end, and flushes
+// to all its children. Example: `
+// ```c++
+// BroadCastLogger({&event_logger1,&event_logger2});
+// ```
+class BroadcastLogger : public EventLogger {
+ public:
+  BroadcastLogger(std::vector<EventLogger *> loggers) : loggers_(loggers) {}
+
+  void AddEvent(Event *event) override {
+    for (EventLogger *logger : loggers_) logger->AddEvent(event);
+  }
+
+  void StartLog() override {
+    for (EventLogger *logger : loggers_) logger->StartLog();
+  }
+
+  void EndLog() override {
+    for (EventLogger *logger : loggers_) logger->EndLog();
+  }
+
+  void Flush() override {
+    for (EventLogger *logger : loggers_) logger->Flush();
+  }
+
+  const std::vector<EventLogger *> &GetLoggers() { return loggers_; }
+
+ private:
+  std::vector<EventLogger *> loggers_;
+};
+
+// A logger for testing `EventLogger`.
+//
+// Mocks the behavior of the virtual methods and provides getters to check if
+// it works as expected.
+class TestLogger : public EventLogger {
+ public:
+  void AddEvent(Event *event) override { events_.push_back(event); }
+
+  void StartLog() override { log_started_ = true; }
+
+  void EndLog() override { log_finished_ = true; }
+
+  void Flush() override { ++flush_count_; }
+
+  const std::vector<Event *> &GetEvents() { return events_; }
+
+  bool IsStarted() const { return log_started_; }
+
+  bool IsFinished() const { return log_finished_; }
+
+  size_t GetFlushCount() const { return flush_count_; }
+
+ private:
+  std::vector<Event *> events_;
+
+  bool log_started_ = false;
+  bool log_finished_ = false;
+  size_t flush_count_ = 0;
+};
 
 TEST(Event, AttributeCreation) {
   const int64_t timestamp_val = 1601314732230797664;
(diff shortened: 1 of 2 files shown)


units/event_log_tests.cc
```diff
@@ -174,7 +175,7 @@ std::string ValueToCSVString(const std::vector<int64_t> &values) {
 
 // Takes an `Event` instance as an input and generates a csv string containing
 // `event`'s name and attribute values.
-// TODO(miladhakimi): Differentiate addresses and other integers. Addresses
+// TODO(miladhakimi): Differentiate hashes and other integers. hashes
```
kuhar: nit: second 'hashes' could be capitalized

units/event_log_tests.cc
```diff
@@ -4,6 +4,7 @@
 #include <string>
 #include <vector>
 
+#include "gmock/gmock.h"
 #include "gtest/gtest.h"
```
kuhar, line 9: You can add `using ::testing::ElementsAre` to simplify the checks

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
   return csv_str.str();
 }
 
+// EventLogger base class
+//
+// EventLogger logs an event in a desired format (CSV, Chrome Trace Event, etc).
```
kuhar: `EventLogger` doesn't know anything about the output format. You can say that the intention is for **concrete** loggers to emit logs in various formats.

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+// It receives a stream of `Event`s as input, converts them to string, and
```
kuhar: `EventLogger` does not perform string conversion

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+// passes them to the LogOutput instance which is resposible for writing the
+// logs. All loggers must implement this class based on their output format
+// specifications.
```
kuhar: > All loggers must implement this class based on their output format
> // specifications.

I don't understand why we mention this. Loggers must implement the interface but that's about what the base class can say about all the other loggers. It's up to the concrete loggers to define what they exactly do

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+class EventLogger {
+ public:
+  virtual ~EventLogger() = default;
+  // Converts an `Event` into a string and writes it to the output.
+  virtual void InsertEvent(Event *event) = 0;
```
kuhar: nit: consider `AddEvent`, it's less typing. Feel free to keep this as-is if you like `Insert` more

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+  // Writes the output format's prefix to the output. For instance, in case of a
+  // CSV log, the the prefix is the CSV header.
```
kuhar: The example is good, but I don't think we should say anything about writing prefix to the output. For example `TestLogger` does not write anything to the output but is a valid implementation. I think it would be better to say that this method is called exactly once before any events are added to denote that log started, and then use the CSV header as example

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+// It receives a stream of `Event`s as input, converts them to string, and
+// passes them to the LogOutput instance which is resposible for writing the
+// logs. All loggers must implement this class based on their output format
+// specifications.
+class EventLogger {
+ public:
+  virtual ~EventLogger() = default;
+  // Converts an `Event` into a string and writes it to the output.
+  virtual void InsertEvent(Event *event) = 0;
+  // Writes the output format's prefix to the output. For instance, in case of a
```
kuhar: nit: Leave an empty line between each function

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+  // CSV log, the the prefix is the CSV header.
+  virtual void StartLog() = 0;
+  // Writes the output format's suffix to the output.
+  virtual void EndLog() = 0;
```
kuhar: same here

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+  // Makes sure all the logs are written to the output.
+  virtual void Flush() = 0;
+};
+
+// FilterLogger is a wrapper around the `EventLogger` which filters the input
```
kuhar: nit: which -> that

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+// events based on a given log level.
+// Example:
+// ``` C++
```
kuhar: there should be no space before c++

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+// TestLogger test;
```
kuhar: nit: I think it's better not to use the `TestLogger` in examples, because real uses outside of tests won't have access to it. Instead, we can write some pseudocode like
```c++
Event* ev = ...; 
FilterLogger logger(ev, LogLevel::High);
```

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+
+  void StartLog() override { logger_->StartLog(); }
+
+  void EndLog() override { logger_->EndLog(); }
+
+  void Flush() override { logger_->Flush(); }
+
+ private:
+  EventLogger *logger_;
+  LogLevel log_level_;
```
kuhar: I realize these fields are initialized in the constructor, but could we also assign some default values here?
```c++
EventLogger *logger_ = nullptr`;
LogLevel ...
```

This is so that we never forget to initialize the fields if we add more constructors in the future.

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+};
+
+// A subclass of `EventLogger` that contains more than one logger and invokes
```
kuhar: `BroadcastLogger` logger can contain one or even zero loggers just fine.

I would say that `BroadCast` logger is an event logger that forwards all events, start/end, and flushes to all its child loggers.

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+    for (EventLogger *logger : loggers_) logger->StartLog();
+  }
+  void EndLog() override {
+    for (EventLogger *logger : loggers_) logger->EndLog();
+  }
+  void Flush() override {
+    for (EventLogger *logger : loggers_) logger->Flush();
+  }
+
+  std::vector<EventLogger *> &GetLoggers() { return loggers_; }
```
kuhar: Do we ever need the non-const version of this function? Can we add `const` like with `GetEvents` in `TestLogger`?

units/event_log_tests.cc
```diff
@@ -205,6 +206,104 @@ std::string EventToCSVString(Event &event) {
...
+  std::vector<EventLogger *> loggers_;
+};
+
+// A logger for testing `EventLogger`.
+//
+// Mocks the behavior of the virtual methods and provides getters to check if
+// it works as expected.
+class TestLogger : public EventLogger {
+ public:
+  TestLogger() : log_started_(false), log_finished_(false), flush_count_(0) {}
```
kuhar: Initialize the fields at definition and remove this constructor.

units/event_log_tests.cc
```diff
@@ -274,4 +373,122 @@ TEST(Event, CreateGraphicsPipelinesEventCreation) {
 
   EXPECT_EQ(EventToCSVString(pipeline_event), expected_str.str());
 }
+
+TEST(EventLogger, MockObjectCreation) {
```
kuhar: MockObjectCreation -> TestLoggerCreation

units/event_log_tests.cc
```diff
@@ -274,4 +373,122 @@ TEST(Event, CreateGraphicsPipelinesEventCreation) {
...
+  TestLogger test_logger;
+  EXPECT_FALSE(test_logger.IsStarted());
+  EXPECT_EQ(0, test_logger.GetFlushCount());
+  EXPECT_FALSE(test_logger.IsFinished());
+  EXPECT_THAT(test_logger.GetEvents(), testing::IsEmpty());
+}
+
+TEST(EventLogger, MockObjectFunctionCalls) {
```
kuhar: MockObject -> TestLogger

units/event_log_tests.cc
```diff
@@ -274,4 +373,122 @@ TEST(Event, CreateGraphicsPipelinesEventCreation) {
...
+  const int64_t timestamp_val = 1601314732230797664;
+  const int64_t hash_val1 = 0x67d6fd0aaa78a6d8;
+  const int64_t hash_val2 = 0x67d390249c2f20ce;
+  const int64_t duration = 926318;
```
kuhar: Because exact numbers don't matter much in these tests, you can pick some short constants like 1, 2, 3, etc. and use them directly without local variables

kuhar: Same in the tests below

units/event_log_tests.cc
```diff
@@ -274,4 +373,122 @@ TEST(Event, CreateGraphicsPipelinesEventCreation) {
...
+
+  broadcast1.Flush();
+  EXPECT_EQ(test_logger1.GetFlushCount(), 1);
+  EXPECT_EQ(test_logger2.GetFlushCount(), 1);
+  EXPECT_EQ(test_logger3.GetFlushCount(), 0);
+
+  broadcast2.Flush();
+  EXPECT_EQ(test_logger1.GetFlushCount(), 2);
+  EXPECT_EQ(test_logger2.GetFlushCount(), 2);
+  EXPECT_EQ(test_logger3.GetFlushCount(), 1);
```
kuhar, line 479: Very cool test setup!

kuhar (changes requested):
Looks great, I only pointed out some cosmetic issues

units/event_log_tests.cc
miladHakimi: Thanks!

units/event_log_tests.cc
```diff
@@ -205,6 +208,115 @@ std::string EventToCSVString(Event &event) {
   return csv_str.str();
 }
 
+// EventLogger base class
+//
+// EventLogger provides an abstraction for the concrete loggers that log the
+// events in various formats (CSV, Chrome Trace Event, etc).
```
kuhar: Maybe mention that event loggers are expected to be internally synchronized?

units/event_log_tests.cc
```diff
@@ -205,6 +208,115 @@ std::string EventToCSVString(Event &event) {
...
+  // Makes sure all the logs are written to the output.
+  virtual void Flush() = 0;
+};
+
+// FilterLogger is a wrapper around the `EventLogger` that filters the input
+// events based on a given log level.
+// Example:
+// ```C++
+// CSVLogger logger = ...;
+// FilterLogger(&logger, LogLevel::High);
```
kuhar: nit: add a variable name for FileLogger

units/event_log_tests.cc
```diff
@@ -205,6 +208,115 @@ std::string EventToCSVString(Event &event) {
...
+  void Flush() override { logger_->Flush(); }
+
+ private:
+  EventLogger *logger_ = nullptr;
+  LogLevel log_level_ = LogLevel::Low;
+};
+
+// A subclass of `EventLogger` that forwards all events, start/end, and flushes
+// to all its children. Example: `BroadCastLogger({&event_logger1,
+// &event_logger2});`
```
kuhar: nit: put this example in a code block, it's a bit difficult to read when it spans 2 lines like this

units/event_log_tests.cc
```diff
@@ -4,9 +4,12 @@
 #include <string>
 #include <vector>
 
+#include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
 namespace {
+using ::testing::ElementsAre;
+using ::testing::IsEmpty;
```
kuhar, line 11: move these out the anonymous namesapce

kuhar (approved):
LGTM

kuhar (approved):

PR #85, opened 2022-10-18
Refactor logging.cc and logging.h to have more special names.

miladHakimi:
The logging.cc and logging.h files contain functions used for the debug logs. That is why their name is changed to debug_logging.The `EventLogger` is also used for logging and will be added to this directory. Using specific file names that describe the functionality of each logger helps us differentiate them easily.


kuhar (approved):
Thank you!

Issue #69, opened 2022-09-26
Refactor data logging


kuhar:
This is a high-level issue for better data logging. Future issues should split this into smaller sub-problems.

performance-layers output two types of logs: event/data logs (`layer_data->Log()`) and debug logs (`LOG(WARNING) << `). We would like to refactor the former one such that layers do not have to worry about the exact output file and output format, yet be able to select from multiple ones. 

The goal is to support future logging formats, e.g., chrome event traces.

Requirements:
- The new classes should be internally-synchronized and not require any mutexes.
- We should be able to unit-test the implementation without having to run the layers.
- Do not introduce any new abseil code.

The final set of classes may look something like this:

```mermaid
graph TD;
  aa(Attribute);
  Int64Attr-->aa;
  StringAttr-->aa;
  DurationAttr-->aa;
  TimestampAttr-->aa;
  ...Atr -->aa;

  ae(Event);
  aa -.->|Contains one or more| ae;
```

```mermaid
graph LR;
  ae(Event);
  EventA-->ae;
  EventB-->ae;
  EventC-->ae;

  el(EventLogger);
  cl(LeafLogger);
  ae-.->|AddEvent| el;
  el-->BroadcastLogger;
  el-->FilterLogger;
  el-->cl;
```

```mermaid
graph TD;
  cl(LeafLogger);
  TestLogger-->cl;
  CsvLogger-->cl;
  CommonLogger-->cl;
  TraceEventLogger-->cl;
```

```mermaid
graph LR;
  el(EventLogger);
  ao(LogOutput);
  el-.->|LogLine| ao;
  ao --> StringOutput & FileOutput;
```

Subtasks:

- [x] https://github.com/googlestadia/performance-layers/issues/75
- [x]  https://github.com/google/vulkan-performance-layers/issues/81
   - [x] Use new logging abstractions for CSV logs
   - [x] Use new logging abstractions for common logs
   - [x] Introduce log output abstraction and implement it for files and strings
- [x] Allow for string log outputs to be used in unit tests
- [x] Add event logging implementation for the Chrome Event Tracing format

kuhar:
Updated the issue to match the updated design.

miladHakimi:
Can you please update this issue by checking the finished subtasks? i think the only remaining subtask is the last one. Right?

miladHakimi:
@kuhar, should we close this issue?

PR #88, opened 2022-10-21
Add implementation for `CSVLogger`.

miladHakimi:
Some events produced by the layers should be written in CSV. `CSVLogger` is an implementation of `EventLogger` that takes events and generates logs in CSV format. It receives the CSV header in its constructor and writes it to the output when `StartLog()` is called. On `AddEvent()`, it converts the incoming event to a CSV string(a line of text) and writes it to the output.
Issues: #81 , #69

### DIFF
--- a/layer/csv_logging.h
+++ b/layer/csv_logging.h
@@ -1,0 +1,85 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_CSV_LOGGING_H_
+#define STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_CSV_LOGGING_H_
+
+#include <cassert>
+#include <cstdio>
+#include <string>
+
+#include "event_logging.h"
+#include "layer_data.h"
+
+namespace performancelayers {
+// Takes an `Event` instance as an input and generates a csv string containing
+// `event`'s name and attribute values.
+// TODO(miladhakimi): Differentiate hashes and other integers. Hashes
+// should be displayed in hex.
+std::string EventToCSVString(Event &event);
+
+// CSVLogger logs the events in the CSV format to the output given in its
+// constructor. Sample use:
+// ```c++
+// CSVLogger logger("pipeline,duration", "compile_time.csv");
+// logger.StartLog();
+// Event compile_time_event = ...;
+// logger.AddEvent(&compile_time_event);
+// logger.Flush();
+// logger.EndLog();
+// ```
+// There is no need to add '\n' at the end of the csv_header in the constructor.
+// This is handled by the implementation. `filename` can be nullptr. In this
+// case, the output will be written to stderr.
+// The only valid methods after calling `EndLog()` are `EndLog()` and the
+// deconstructor.
+class CSVLogger : public EventLogger {
+ public:
+  CSVLogger(const char *csv_header, const char *filename);
+
+  ~CSVLogger() {
+    if (out_ && out_ != stderr) fclose(out_);
+  }
+
+  void AddEvent(Event *event) override {
+    assert(out_);
+    std::string event_str = EventToCSVString(*event);
+    WriteLnAndFlush(out_, event_str);
+  }
+
+  // Writes the CSV header given in the constructor to the output.
+  void StartLog() override {
+    assert(out_);
+    WriteLnAndFlush(out_, header_);
+  }
+
+  void EndLog() override {
+    if (out_ && out_ != stderr) {
+      fclose(out_);
+      out_ = nullptr;
+    }
+  }
+
+  void Flush() override {
+    assert(out_);
+    fflush(out_);
+  }
+
+ private:
+  FILE *out_ = nullptr;
+  const char *header_ = nullptr;
+};
+
+}  // namespace performancelayers
+#endif  // STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_CSV_LOGGING_H_
--- a/units/csv_log_tests.cc
+++ b/units/csv_log_tests.cc
@@ -1,0 +1,39 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "csv_logging.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+namespace performancelayers {
+namespace {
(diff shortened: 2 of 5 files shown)


layer/csv_logging.h
```diff
@@ -0,0 +1,76 @@
...
+  // Writes the CSV header given in the constructor to the output.
+  void StartLog() override { WriteLnAndFlush(out_, header_); }
+
+  void EndLog() override {}
+
+  void Flush() override { fflush(out_); }
+
+ private:
+  FILE *out_ = nullptr;
+  const char *header_;
```
kuhar: nit `header_ = nullptr;`

layer/csv_logging.h
```diff
@@ -0,0 +1,76 @@
...
+      : header_(csv_header) {
+    if (filename) {
+      out_ = fopen(filename, "w");
+    } else {
+      out_ = stderr;
+    }
+  }
+
+  ~CSVLogger() {
+    if (out_ != stderr) fclose(out_);
```
kuhar: Shouldn't we close the file on `EndLog` and reset the handle? Here we could conditionally close it if `OnEnd` wasn't called

layer/csv_logging.h
```diff
@@ -0,0 +1,76 @@
...
+// Event compile_time_event = ...;
+// logger.AddEvent(&compile_time_event);
+// logger.Flush();
+// logger.EndLog();
+// ```
+// There is no need to add '\n' at the end of the csv_header in the constructor.
+// This is handled by the implementation.
+class CSVLogger : public EventLogger {
+ public:
+  CSVLogger(const char *csv_header, const char *filename)
```
kuhar: Can you add a short comment explaining what happens when filename is a nullptr?

layer/csv_logging.h
```diff
@@ -0,0 +1,76 @@
...
+      : header_(csv_header) {
+    if (filename) {
+      out_ = fopen(filename, "w");
```
kuhar: What happens when `fopen` fails? E.g., end user provides a bad path

kuhar (changes requested):

layer/csv_logging.h
miladHakimi: Since we're opening it in the constructor and once, we can only close it on the destructor. If I have it on `EndLog()`, I also need to check if the file is closed  in the destructor and close it if it isn't (the `EndLog()` was not called). I also don't need to worry about multiple `EndLog()` invocations.

layer/csv_logging.h
miladHakimi: Sure!

layer/csv_logging.h
kuhar: > Since we're opening it in the constructor and once, we can only close it on the destructor.

Why? This is what `std::fstream` does on `.close()`: https://en.cppreference.com/w/cpp/io/basic_fstream

> I also need to check if the file is closed in the destructor and close it if it isn't

We can set the file handle to nullptr and close only when not null and not stderr.

kuhar: > I also don't need to worry about multiple EndLog() invocations.

I see your point but I think it would be easier to just make `EndLog` support this scenario. From the practical perspective it may be convenient to make double EndLog a noop and only disallow adding event after end

miladHakimi: > > Since we're opening it in the constructor and once, we can only close it on the destructor.
> 
> Why? This is what `std::fstream` does on `.close()`: https://en.cppreference.com/w/cpp/io/basic_fstream

Cool! I didn't know it closes the file in the destructor.

miladHakimi: > We can set the file handle to nullptr and close only when not null and not stderr.

Yep, That's what I did first then I changed it to this. I'll fix it.

layer/csv_logging.h
miladHakimi: go/cstyle#Doing_Work_in_Constructors talks briefly about error handling in constructors. There are a couple of options to deal with `fopen` failure:
1. Using stderr instead.
2. exiting the program.
3. Taking the file descriptor as an argument. In this case, the user is responsible for handling the `fopen` failure.
4. Setting an `is_valid` flag and have a public getter for it. The user should check if the logger is valid before using the methods.

kuhar: Continuing with stderr sounds fine to me. This way the user can know the layer is working but may be misconfigured

kuhar: We could also print an error message to stderr when the filename was not provided or was bad

kuhar: (Using SPL_LOG)

layer/csv_logging.h
```diff
@@ -0,0 +1,94 @@
...
+                       << ". Using stderr as the alternative output.";
+        out_ = stderr;
+      }
+    } else {
+      out_ = stderr;
+    }
+  }
+
+  ~CSVLogger() {
+    if (out_ != stderr) fclose(out_);
```
kuhar: What if `out_` is a nullptr? `fclose(nullptr)` is UB

layer/csv_logging.h
```diff
@@ -0,0 +1,94 @@
...
+// There is no need to add '\n' at the end of the csv_header in the constructor.
+// This is handled by the implementation. `filename` can be nullptr. In this
+// case, the output will be written to stderr.
+class CSVLogger : public EventLogger {
+ public:
+  CSVLogger(const char *csv_header, const char *filename)
+      : header_(csv_header) {
+    if (filename) {
+      out_ = fopen(filename, "w");
+      if (out_ == nullptr) {
```
kuhar: nit: `if (!out_)`

layer/csv_logging.h
```diff
@@ -0,0 +1,94 @@
...
+      out_ = stderr;
+    }
+  }
+
+  ~CSVLogger() {
+    if (out_ != stderr) fclose(out_);
+  }
+
+  void AddEvent(Event *event) override {
+    if (out_) {
```
kuhar: We can say that the only valid thing to do after `EndLog` is to call it again or run the destructor. And use `assert(out_)` here instead.

layer/csv_logging.h
```diff
@@ -0,0 +1,94 @@
...
+      std::string event_str = EventToCSVString(*event);
+      WriteLnAndFlush(out_, event_str);
+    }
+  }
+
+  // Writes the CSV header given in the constructor to the output.
+  void StartLog() override {
+    if (out_) WriteLnAndFlush(out_, header_);
```
kuhar: also here

layer/csv_logging.h
```diff
@@ -0,0 +1,94 @@
...
+
+  void EndLog() override {
+    if (out_ && out_ != stderr) {
+      fclose(out_);
+      out_ = nullptr;
+    }
+  }
+
+  void Flush() override {
+    if (out_) fflush(out_);
```
kuhar: also here

layer/csv_logging.h
```diff
@@ -0,0 +1,94 @@
...
+// logger.AddEvent(&compile_time_event);
+// logger.Flush();
+// logger.EndLog();
+// ```
+// There is no need to add '\n' at the end of the csv_header in the constructor.
+// This is handled by the implementation. `filename` can be nullptr. In this
+// case, the output will be written to stderr.
+class CSVLogger : public EventLogger {
+ public:
+  CSVLogger(const char *csv_header, const char *filename)
```
kuhar: This constructor is getting long, we should move it to the `.cc` file.

kuhar (changes requested):

layer/csv_logging.h
miladHakimi: If I use assert(out_) then calling `EndLog` after `EndLog` is not ok. right?

units/csv_log_tests.cc
```diff
@@ -0,0 +1,37 @@
...
+TEST(CSVLogger, MethodCheck) {
+  CSVLogger logger("name,timestamp,pipeline,duration", nullptr);
+  VectorInt64Attr hashes("hashes", {2, 3});
+  CreateGraphicsPipelinesEvent pipeline_event("create_graphics_pipeline", 1,
+                                              hashes, 4, LogLevel::kHigh);
+  logger.StartLog();
+  logger.AddEvent(&pipeline_event);
+  logger.Flush();
+  logger.EndLog();
+
```
kuhar, line 34: Since we support the double EndLog scenario, we should have a testcase that exercises it.

kuhar (approved):
LGTM

cassiebeckley (approved):
LGTM

PR #91, opened 2022-10-25
Add `MemoryUsageEvent` to memory usage layer.

miladHakimi:
This PR implements the `MemoryUsageEvent` class. It is an event that contains the allocation information such as current and peak allocated bytes. This event should be logged to both private and the common files. It will be used in both `QueuePresentKHR()` and `DestroyDevice()` instead of the current `LogUsage()` call.

### DIFF
--- a/layer/memory_usage_layer.cc
+++ b/layer/memory_usage_layer.cc
@@ -12,10 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cstddef>
+#include <cstdint>
+#include <vector>
+
 #include "debug_logging.h"
+#include "event_logging.h"
 #include "layer_data.h"
 #include "layer_utils.h"
 
+namespace performancelayers {
 namespace {
 // ----------------------------------------------------------------------------
 // Layer book-keeping information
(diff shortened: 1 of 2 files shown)


layer/memory_usage_layer.cc
```diff
@@ -12,10 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <stddef.h>
```
kuhar: Prefer C++ versions of C headers, here `cstddef`

layer/memory_usage_layer.cc
```diff
@@ -12,10 +12,17 @@
...
+
```
kuhar: nit: delete the empty line between system headers

layer/memory_usage_layer.cc
```diff
@@ -26,7 +33,29 @@ constexpr char kLayerName[] = "VK_LAYER_STADIA_memory_usage";
...
+      : Event(name, LogLevel::kHigh),
+        current_({"current", current}),
+        peak_({"peak", peak}),
+        attributes_{&current_, &peak_} {}
+
+  const std::vector<Attribute*>& GetAttributes() override {
+    return attributes_;
+  }
+
+  size_t GetNumAttributes() const override { return attributes_.size(); }
```
kuhar: What do you think about adding a default implementation to `Event` that does `GetAttributes().size()`?

kuhar (commented):
Could we use the CSVLogger here to output this event? If not, what is missing?

miladHakimi:
> Could

Yes we can. Actually, it's already on my repo. I just wanted to keep this PR small.

kuhar:
> > Could
> 
> Yes we can. Actually, it's already on my repo. I just wanted to keep this PR small.

I'd prefer any new code to be used, either in tests or by layers. Unused code tends to bitrot surprisingly quickly.

miladHakimi:
> I'd prefer any new code to be used, either in tests or by layers. Unused code tends to bitrot surprisingly quickly.

I can make it one PR with multiple commits.

PR #92, opened 2022-10-25
Rafactor memory usage layer to use CSVLogger.

miladHakimi:
Add `MemoryUsageEvent` to memory usage layer.

This PR implements the `MemoryUsageEvent` class. It is an event that contains the allocation information such as current and peak allocated bytes. The data should be logged to both private and the common files. `MemoryUsageEvent` is used in both `QueuePresentKHR()` and `DestroyDevice()` instead of the current `LogUsage()` call.

`MemoryUsageLayerData` allocates and deallocates The `CSVLogger`. The logger accepts events through the new `LogEvent` method. `LogEvent` takes a `MemoryUsageEvent` as input and logs it to the memory layer's log file.

### DIFF
--- a/layer/memory_usage_layer.cc
+++ b/layer/memory_usage_layer.cc
@@ -12,10 +12,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <algorithm>
+#include <cstddef>
+#include <cstdint>
+#include <vector>
+
+#include "csv_logging.h"
 #include "debug_logging.h"
+#include "event_logging.h"
 #include "layer_data.h"
 #include "layer_utils.h"
 
+namespace performancelayers {
 namespace {
 // ----------------------------------------------------------------------------
 // Layer book-keeping information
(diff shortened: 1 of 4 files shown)


layer/memory_usage_layer.cc
```diff
@@ -86,6 +132,8 @@ class MemoryUsageLayerData : public performancelayers::LayerData {
 
   VkDeviceSize current_allocation_size_ ABSL_GUARDED_BY(memory_hash_lock_) = 0;
   VkDeviceSize peak_allocation_size_ ABSL_GUARDED_BY(memory_hash_lock_) = 0;
+  CSVLogger* private_logger_;
+  FilterLogger* private_logger_filter_;
```
kuhar: We should use values instead of pointers if we can.

layer/memory_usage_layer.cc
```diff
@@ -26,13 +34,47 @@ constexpr char kLayerName[] = "VK_LAYER_STADIA_memory_usage";
...
+      : Event(name, LogLevel::kHigh),
+        current_({"current", current}),
+        peak_({"peak", peak}),
+        attributes_{&current_, &peak_} {}
+
+  const std::vector<Attribute*>& GetAttributes() override {
+    return attributes_;
+  }
+
+  size_t GetNumAttributes() const override { return attributes_.size(); }
```
kuhar: Can we make this the default implementation for all `Event`s? I expect all of them to implement it the same way anyway.

kuhar (commented):
Looks great, just one nit

layer/memory_usage_layer.cc
miladHakimi: Yes. Sure.

kuhar (approved):
awesome!

PR #93, opened 2022-10-26
Implement `LayerDataWithLogger` to add logger to `LayerData`.

miladHakimi:
Currently, to have loggers in the memory layer, we add them to the MemoryUsageLayerData which extends the LayerData. To Add loggers in the same way to the other layers, we need to apply the same changes to them which leads to duplicate code. Since there are layers that don't need loggers, we don't add the loggers to the LayerData class.

LayerDataWithLogger extends the LayerData by adding a CSV logger. The layers that require a logger can use it instead of LayerData.

This PR also refactors the memory usage layer to use the new class.

### DIFF
--- a/layer/common_logging.h
+++ b/layer/common_logging.h
@@ -20,7 +20,7 @@
 #include <string>
 
 #include "event_logging.h"
-#include "layer_data.h"
+#include "layer_utils.h"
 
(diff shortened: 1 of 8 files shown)


kuhar (approved):
LGTM

PR #94, opened 2022-10-27
Refactor the code to use chrono time instead of absl.

miladHakimi:
One of the future directions of perfomancelayers is to have dynamic
unloading to provide better support for the applications. Currently, the
`absl` libraries we're using don't support dynamic unloading. Hence, we
should not use `absl` in the new features and refactor the current
codebase to use alternative libraries.

This PR refactors the performancelayers to use `chrono` time instead of
`absl` time.

`chrono::steady_clock`, `chrono::system_clock`,
`chrono::high_resolution_clock`, and `absl::Now()` brief explanation:
1. `steady_clock`: A monotonic clock. The current value of steady_clock
does not matter, but the guarantee that is strictly increasing is useful
for calculating the difference between two samples.
2. `system_clock`: Is the wall-clock. Used for time/day related stuff.
3. `high_resolution_clock`: An alias for one of the above.
4. `absl::Now()`: Constructs an absl::Time from the system clock.

Previously, `absl::Now()` was used for the timestamp and measuring
duration. This PR separates these two by using `system_clock::now()` for
the timestamp and `steady_clock::now()` for the duration.

### DIFF
--- a/layer/compile_time_layer.cc
+++ b/layer/compile_time_layer.cc
@@ -23,6 +23,7 @@
 #include "layer_data.h"
 #include "layer_utils.h"
 
+namespace performancelayers {
 namespace {
 // ----------------------------------------------------------------------------
 // Layer book-keeping information
@@ -34,7 +35,7 @@ constexpr char kLayerDescription[] =
     "Stadia Pipeline Compile Time Measuring Layer";
 constexpr char kLogFilenameEnvVar[] = "VK_COMPILE_TIME_LOG";
 
-class CompileTimeLayerData : public performancelayers::LayerData {
+class CompileTimeLayerData : public LayerData {
  public:
   CompileTimeLayerData(char* log_filename)
       : LayerData(log_filename, "Pipeline,Compile Time (ns)") {
@@ -44,13 +45,13 @@ class CompileTimeLayerData : public performancelayers::LayerData {
   // Used to track the slack between shader module creation and its first use in
   // pipeline creation.
   struct ShaderModuleSlack {
-    std::optional<absl::Time> creation_end_time = std::nullopt;
-    std::optional<absl::Time> first_use_time = std::nullopt;
+    std::optional<DurationClock::time_point> creation_end_time = std::nullopt;
+    std::optional<DurationClock::time_point> first_use_time = std::nullopt;
   };
 
   // Records the time |create_end| of the |shader| creation.
   void RecordShaderModuleCreation(VkShaderModule shader,
-                                  absl::Time create_end) {
+                                  DurationClock::time_point create_end) {
     absl::MutexLock lock(&shader_module_usage_lock_);
     assert(!shader_module_to_usage_.contains(shader) &&
            "Shader already created");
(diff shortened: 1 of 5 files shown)


layer/compile_time_layer.cc
```diff
@@ -28,6 +29,12 @@ namespace {
 // Layer book-keeping information
 // ----------------------------------------------------------------------------
 
+using time_point = std::chrono::system_clock::time_point;
```
kuhar: Can you add a comment explaining why we want to use this particular clock instead `steady_clock` or `high_resolution_clock`?

layer/compile_time_layer.cc
```diff
@@ -28,6 +29,12 @@ namespace {
...
+
+// A function alias used to get the current time_point from chrono and replaces
+// absl::now().
+auto Now = std::chrono::system_clock::now;
```
kuhar: nit: `constexpr auto Now`? Or make it a function, this will make it easier to tell what the return type is and it's not much more typing:
```
time_point Now() { return ...; }
```

layer/compile_time_layer.cc
```diff
@@ -164,11 +171,11 @@ SPL_COMPILE_TIME_LAYER_FUNC(VkResult, CreateComputePipelines,
   assert(create_info_count > 0 &&
          "Specification says create_info_count must be > 0.");
 
-  absl::Time start = absl::Now();
+  auto start = Now();
```
kuhar: We should spell out the exact type because it's not obvious by just looking at the RHS. Also everywhere else,

layer/layer_data.cc
```diff
@@ -25,6 +25,11 @@
 #include "layer_utils.h"
 
 namespace {
+// A function alias used to get the current time_point from chrono and replaces
+// absl::now().
+auto Now = std::chrono::system_clock::now;
+
+using time_point = std::chrono::system_clock::time_point;
```
kuhar: If all layers are supposed to use the same clock and `Now` function, we should move it to layer_utils.

kuhar (changes requested):

layer/layer_utils.h
```diff
@@ -53,6 +55,28 @@
...
+// not matter, but the guarantee that is strictly increasing is useful for
+// calculating the difference between two samples.
+// 2. system_clock: Is the wall-clock. Used for time/day related stuff.
+// 3. high_resolution_clock: An alias for one of the above.
+
+// Returns a system_clock::time_point as the timestamp. The system_clock acts
+// like a real clock and shows the current time, while the steady_clock is a
+// monotonic clock and behaves like a chronometer. This is why system_clock is
+// used for the timestamp.
+std::chrono::system_clock::time_point GetTimestamp();
```
kuhar: Consider introducing aliases for these clocks here that layers can use.
```c++
using TimestampClock = std::chrono::system_clock;
using DurationClock = std::chrono::steady_lock;
```

I think this would help ensuring that layers use the appropriate clocks.

layer/compile_time_layer.cc
```diff
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 #include <cassert>
+#include <chrono>
```
kuhar: nit: no need to include chrono here, because the file does not directly name anything in chrono. The type behind `DurationClock` will introduce the necessary includes.

layer/layer_data.h
```diff
@@ -15,6 +15,7 @@
 #ifndef STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_LAYER_DATA_H_
 #define STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_LAYER_DATA_H_
 
+#include <chrono>
```
kuhar: same here

kuhar (commented):
Nice, just one nit.

layer/layer_data.h
miladHakimi: `layer_data .h` is using `time_point::min()`. Doesn't that count?

kuhar: no

kuhar (approved):
LGTM

PR #95, opened 2022-10-27
Refactor frame time layer to use `CSVLogger`.

miladHakimi:
This PR adds support for bool type attributes and implements the FrameTimeEvent.

One side effect of this refactor is that it increases the performance of the frame time layer by reducing the critical section of the multi-threaded region. This is achieved by writing the log after updating the shared variable, last_time_log_.
last_time_log_ is a shared variable protected by a mutex. Previously, the LogTimeDelta() acquired the mutex, computed the time delta, logged an event based on time delta, and updated the last_time_log_. However, updating last_time_log_ could be done right after computing time delta without having to wait for the event to be written to the output.

GetTimeDelta() is a new function responsible for computing the time delta and updating the last_time_log_. The layer uses the time delta returned by this function and creates the log based on that outside of the critical section.


kuhar (approved):
LGTM

PR #96, opened 2022-10-31
Refactor compile time layer to use `CSVLogger`.

miladHakimi:
This PR implements `CompileTimeEvent` which is an event that holds hash values and pipeline creation duration. This event is used by `CreateComputePipelines()` and `CreateGraphicsPipelines()` functions.

### DIFF
--- a/layer/compile_time_layer.cc
+++ b/layer/compile_time_layer.cc
@@ -17,9 +17,12 @@
 #include <cstdlib>
 #include <cstring>
 #include <functional>
+#include <iomanip>
 #include <optional>
+#include <sstream>
 #include <string>
 
+#include "event_logging.h"
 #include "layer_data.h"
 #include "layer_utils.h"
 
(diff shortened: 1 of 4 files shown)


layer/compile_time_layer.cc
```diff
@@ -35,10 +36,23 @@ constexpr char kLayerDescription[] =
     "Stadia Pipeline Compile Time Measuring Layer";
 constexpr char kLogFilenameEnvVar[] = "VK_COMPILE_TIME_LOG";
 
-class CompileTimeLayerData : public LayerData {
+class CompileTimeEvent : public Event {
+ public:
+  CompileTimeEvent(const char* name, std::vector<int64_t>& hash_values,
+                   int64_t duration)
```
kuhar: Shouldn't we have a separate attribute for duration? E.g., `DurationNsAttr`. This way it's obvious what the unit is.

layer/layer_data.h
```diff
@@ -40,6 +40,12 @@
 #include "vk_layer_dispatch_table.h"
 
 namespace performancelayers {
+// Returns a quoted string.
+template <typename StrTy>
+std::string QuoteStr(StrTy&& str) {
+  return absl::StrCat("\"", std::forward<StrTy>(str), "\"");
```
kuhar: Can we use: `std::quoted`? https://en.cppreference.com/w/cpp/io/manip/quoted

kuhar (changes requested):

layer/event_logging.h
```diff
@@ -89,6 +89,7 @@ using VectorInt64Attr =
 using StringAttr = AttributeImpl<std::string, ValueType::kString>;
 using Int64Attr = AttributeImpl<int64_t, ValueType::kInt64>;
 using BoolAttr = AttributeImpl<bool, ValueType::kBool>;
+using DurationNsAttr = AttributeImpl<int64_t, ValueType::kInt64>;
```
kuhar: We would need a separate ValueType id to represent duration. One way would be to create a new type (struct/class) representing durations and adding an id for it in the `ValueType` enum.

layer/compile_time_layer.cc
```diff
@@ -35,10 +38,25 @@ constexpr char kLayerDescription[] =
     "Stadia Pipeline Compile Time Measuring Layer";
 constexpr char kLogFilenameEnvVar[] = "VK_COMPILE_TIME_LOG";
 
-class CompileTimeLayerData : public LayerData {
+class CompileTimeEvent : public Event {
+ public:
+  CompileTimeEvent(const char* name, std::vector<int64_t>& hash_values,
```
kuhar: make `hash_values` a const ref

layer/layer_data.h
kuhar: Not resolved yet?

kuhar (changes requested):

kuhar (approved):
LGTM

PR #97, opened 2022-11-01
Fix reorder constructor warning/error.

miladHakimi:
The constructor of the base class is always called before the
constructor of the subclass. This is true even if we call it at the end
of the constructor initializer list of the subclass.

We want to initialize `Event`'s `attributes_` with pointers coming from
the subclass upon construction. The subclasses contain different
attributes, create `vector<Attribute*>` in their constructor and
pass it to the `Event` constructor. Since we were passing pointers, the
program was still correct even if pointers were initialized after being
passed to the base constructor. The error says we must move the
call to the `Event` constructor to the beginning of the initializer
list. Despite being correct, initializing the variables after
passing them to the `Event` might look confusing. This is why we defined
a protected method, `InitAttributes()`, that takes the attribute
pointers and builds the `vector<Attribute*>` in the base class.

### DIFF
--- a/layer/event_logging.h
+++ b/layer/event_logging.h
@@ -95,9 +95,8 @@ using BoolAttr = AttributeImpl<bool, ValueType::kBool>;
 // initialize their own set of attributes.
 class Event {
  public:
-  Event(const char *name, std::vector<Attribute *> attributes,
-        LogLevel log_level)
-      : name_(name), attributes_(attributes), log_level_(log_level) {}
+  Event(const char *name, LogLevel log_level)
+      : name_(name), log_level_(log_level) {}
 
   virtual ~Event() = default;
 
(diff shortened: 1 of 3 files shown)


cassiebeckley (approved):
LGTM but I'd like to know if @kuhar knows a cleaner way of accomplishing this

layer/event_logging.h
```diff
@@ -109,9 +108,11 @@ class Event {
 
   LogLevel GetLogLevel() const { return log_level_; }
 
+ protected:
+  std::vector<Attribute *> attributes_;
```
kuhar: Instead of making the variable protected, what do you think about adding a protected function `initAttributes`? This should be more greppable and easier to debug because all updates will be in the same function

layer/event_logging.h
```diff
@@ -109,10 +108,16 @@ class Event {
 
   LogLevel GetLogLevel() const { return log_level_; }
 
+ protected:
+  template <typename... Args>
+  void InitAttributes(Args... args) {
```
kuhar: either `void InitAttributes(Args*... args)` to make it clear that we expect pointers
or alternatively, we can get rid of the template with something like:
```c++
void InitAttributes(std::initializer_list<Attribute *> attrs) {
  attributes_ = {attrs.begin(), attrs.end()};
}
```

and then use like: `InitAttributes({&attr1, &attr2, &attr3})`

kuhar (approved):
LGTM, thanks for taking care of this!

PR #98, opened 2022-11-01
Add support for duration attribute.

miladHakimi:
By using an `Int64Attr` as a single number that represents time-related values (duration/timestamp), we would loose the time unit information.

The `DurationAttr` is the base class for attributes that contain duration information. It is a wrapper around a `chrono::steady_clock::duration` variable. Each subclass implements the `DurationAttr` based on a time unit and converts the duration variable into an `int` to represent it in that unit.

### DIFF
--- a/layer/event_logging.h
+++ b/layer/event_logging.h
@@ -15,18 +15,22 @@
 #ifndef STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_EVENT_LOGGING_H_
 #define STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_EVENT_LOGGING_H_
 
+#include <cstdint>
 #include <string>
 #include <vector>
 
+#include "layer_utils.h"
+
 namespace performancelayers {
 // Set of supported value types for an attribute.
-enum ValueType { kString, kInt64, kVectorInt64, kBool };
+enum ValueType { kString, kInt64, kVectorInt64, kBool, kDuration };
 
 // Specifies the importance of an Eventk. Events are logged based on their level
 // of importance. E.g., compile_time.csv only containsk the important compile
 // time events while event.log contains all kthe compile time events.
 enum class LogLevel { kLow, kMedium, kHigh };
 
+enum class TimeUnit { KNano, kMilli, kSec };
 // Attribute class
 //
 // An event consists of a set of attributes. Each attribute has a name that
@@ -84,6 +88,23 @@ class AttributeImpl : public Attribute {
   T value_;
 };
 
+// The base class for an attribute that keeps the duration information. To take
+// the time unit into account, the subclasses must implement the `GetValue()`
+// method such that it converts the underlying `duration` value to a specific
+// unit(nanoseconds, miliseconds, seconds, etc).
+class DurationAttr : public Attribute {
+ public:
+  static constexpr ValueType id_ = ValueType::kDuration;
+
+  DurationAttr(const char *name, const DurationClock::duration &value)
+      : Attribute(name, ValueType::kDuration), value_(value) {}
+
+  DurationClock::duration GetValue() const { return value_; };
+
+ private:
+  DurationClock::duration value_;
+};
+
 using VectorInt64Attr =
     AttributeImpl<std::vector<int64_t>, ValueType::kVectorInt64>;
 using StringAttr = AttributeImpl<std::string, ValueType::kString>;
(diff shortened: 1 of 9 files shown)


layer/event_logging.h
```diff
@@ -84,6 +87,43 @@ class AttributeImpl : public Attribute {
...
+  virtual int64_t &GetValue() = 0;
+
+  const DurationClock::duration &GetDuration() { return value_; }
+
+ private:
+  DurationClock::duration value_;
+};
+
+// An implementation of `DurationAttr` for `nanoseconds`.
+class DurationNsAttr : public DurationAttr {
```
kuhar: I think DurationNs should the only DurationAttribute. We can then add more getter functions if we need less than nanosecond precision.

layer/event_logging.h
```diff
@@ -84,6 +87,43 @@ class AttributeImpl : public Attribute {
...
 
+// The base class for an attribute that keeps the duration information. To take
+// the time unit into account, the subclasses must implement the `GetValue()`
+// method such that it converts the underlying `duration` value to a specific
+// unit(nanoseconds, miliseconds, seconds, etc).
+class DurationAttr : public Attribute {
+ public:
+  static constexpr ValueType id_ = ValueType::kDuration;
+
+  virtual const int64_t &GetValue() const = 0;
```
kuhar: Why not return the underlying duration?

layer/event_logging.h
```diff
@@ -84,6 +87,43 @@ class AttributeImpl : public Attribute {
...
+
+ protected:
+  DurationAttr(const char *name, const DurationClock::duration &value)
+      : Attribute(name, ValueType::kDuration), value_(value) {}
+
+  virtual int64_t &GetValue() = 0;
```
kuhar: Do we need to return the ref? Could we make it immutable instead and return it by value?

layer/event_logging.h
kuhar: And additionally, we can add more getter to get the number of nanoseconds etc.

miladHakimi: > Why not return the underlying duration?

I wanted to treat the value returned by this function as a value that doesn't need an extra step of conversion so that the logger doesn't need to be concerned about the time unit.

kuhar: Oh, I see. So you want the dynamic type of the duration to determine the printing format. In this case, you would need a new ID for each duration unit.

kuhar: Edit: I think I get it know, now, this is what's already going on, except that DurationNsAttr doesn't have it's own unit.
This seems fine to me then, just return `int64_t` by value.

layer/event_logging.h
```diff
@@ -84,6 +87,43 @@ class AttributeImpl : public Attribute {
...
+  DurationNsAttr(const char *name, const DurationClock::duration &value)
+      : DurationAttr(name, value),
+        int_value_(ToInt64Nanoseconds(GetDuration())) {}
+  const int64_t &GetValue() const override { return int_value_; }
+
+ protected:
+  int64_t &GetValue() override { return int_value_; }
+
+ private:
+  int64_t int_value_;
```
kuhar: We don't need this member. You can call the helper function to convert duration to nannos.

layer/event_logging.h
```diff
@@ -84,6 +87,43 @@ class AttributeImpl : public Attribute {
...
+// An implementation of `DurationAttr` for `nanoseconds`.
+class DurationNsAttr : public DurationAttr {
+ public:
+  DurationNsAttr(const char *name, const DurationClock::duration &value)
+      : DurationAttr(name, value),
+        int_value_(ToInt64Nanoseconds(GetDuration())) {}
+  const int64_t &GetValue() const override { return int_value_; }
+
+ protected:
+  int64_t &GetValue() override { return int_value_; }
```
kuhar: don't need this one

kuhar (changes requested):

layer/event_logging.h
```diff
@@ -84,6 +87,43 @@ class AttributeImpl : public Attribute {
...
+
+  virtual const int64_t &GetValue() const = 0;
+
+ protected:
+  DurationAttr(const char *name, const DurationClock::duration &value)
+      : Attribute(name, ValueType::kDuration), value_(value) {}
+
+  virtual int64_t &GetValue() = 0;
+
+  const DurationClock::duration &GetDuration() { return value_; }
```
kuhar: This can be public

kuhar (approved):

PR #100, opened 2022-11-02
Add support for timestamp attribute.

miladHakimi:
By using an Int64Attr as a single number that represents time-related values (duration/timestamp), we would loose the time unit information.

The `TimestampAttr` is an attribute that contains timestamp information. It is a wrapper around a `chrono::system_clock::time_point` variable. This variable has nanoseconds precision and the loggers are responsible for converting it to the desired time unit.

### DIFF
--- a/layer/event_logging.h
+++ b/layer/event_logging.h
@@ -23,14 +23,13 @@
 
 namespace performancelayers {
 // Set of supported value types for an attribute.
-enum ValueType { kString, kInt64, kVectorInt64, kBool, kDuration };
+enum ValueType { kBool, kDuration, kInt64, kString, kTimestamp, kVectorInt64 };
 
 // Specifies the importance of an Eventk. Events are logged based on their level
 // of importance. E.g., compile_time.csv only containsk the important compile
 // time events while event.log contains all kthe compile time events.
 enum class LogLevel { kLow, kMedium, kHigh };
 
-enum class TimeUnit { KNano, kMilli, kSec };
 // Attribute class
 //
 // An event consists of a set of attributes. Each attribute has a name that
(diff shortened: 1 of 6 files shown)


layer/event_logging.h
```diff
@@ -23,7 +23,7 @@
 
 namespace performancelayers {
 // Set of supported value types for an attribute.
-enum ValueType { kString, kInt64, kVectorInt64, kBool, kDuration };
+enum ValueType { kString, kInt64, kVectorInt64, kBool, kDuration, kTimestamp };
```
kuhar: nit: can you put these in the lexicographic order? This will make it easier to quickly scan looking for some specific name

units/event_log_tests.cc
```diff
@@ -69,7 +69,7 @@ TEST(Event, AttributeCreation) {
 }
 
 TEST(Event, CreateShaderModuleEventCreation) {
-  const int64_t timestamp_val = 1601314732230797664;
+  TimestampClock::time_point timestamp_val;
```
kuhar: Why isn't this initialized? What's the default value?

kuhar: Also below

kuhar (changes requested):
Looks great, just two things that stood out to me

units/event_log_tests.cc
miladHakimi: The default time_point is  00:00:00 - 1 January 1970. I can initialize all with `GetTimestamp()` function to be more clear.

kuhar: How about something in the middle: initialize it with the default, whatever it is: `TimestampClock::time_point x = {};`?

miladHakimi: Ok

kuhar (approved):

PR #102, opened 2022-11-03
Enable `warning-as-error` and fix the errors.

miladHakimi:
Some of the current unhandled warnings cause failures in the imports because the imports consider them as errors.

This PR enables warnings in the `cmake` and sets the `warning-as-error` flag in the CI  (except for `comment` and `return-type` warnings`) so that the CI can catch the warnings that would break the imports. It also fixes some existing warnings such as missing initializations, unhandled switch-case statements, and unclosed files.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,13 @@ set(CMAKE_CXX_STANDARD 17)
 # Disable C++ exceptions.
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-exceptions -fno-rtti")
 
+# Enable warnings.
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall -Wextra")
+
+if ( CMAKE_CXX_COMPILER_ID STREQUAL "Clang" )
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wthread-safety")
+endif()
+
 set(FILESYSTEM_LIB_NAME stdc++fs)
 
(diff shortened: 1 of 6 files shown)


CMakeLists.txt
```diff
@@ -22,6 +22,9 @@ set(CMAKE_CXX_STANDARD 17)
 # Disable C++ exceptions.
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fno-exceptions -fno-rtti")
 
+# Set `warning-as-error` for all the warnings except `comment` and `return-type`.
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Werror -Wall -Wextra -Wno-comment -Wno-return-type")
```
kuhar: `Werror` for all of these seems too aggressive to me. Depending on the compiler version, the project can fail to build on somebody else's system even if it passes our CI.

I'd enable the warnings here but move the `-Werror` to the cmake invocation in docker, so that warnings don't break local builds.

units/csv_log_tests.cc
```diff
@@ -23,7 +23,7 @@ namespace {
 TEST(CSVLogger, MethodCheck) {
   CSVLogger logger("name,timestamp,pipeline,duration", nullptr);
   VectorInt64Attr hashes("hashes", {2, 3});
-  DurationClock::duration dur;
+  DurationClock::duration dur(0);
```
kuhar: nit: `dur = 0;` or `dur{};`

kuhar (changes requested):

kuhar (approved):

PR #103, opened 2022-11-03
tests: fix cond jump depends on uninitialised val

Keenuts:
Running tests with valgrind showed this error. Fixing.

Signed-off-by: Nathan Gauër <brioche@google.com>


kuhar (approved):
Thank you!

PR #105, opened 2022-11-03
Refactor runtime layer to use `CSVLogger`.

miladHakimi:
This PR implements RuntimeEvent which is an event that holds hash values, pipline runtime, and the invocation count of the fragmanet and compute shaders. This event is used by `LogAndRemoveQueryPools()`.

### DIFF
--- a/layer/runtime_layer_data.cc
+++ b/layer/runtime_layer_data.cc
@@ -14,9 +14,13 @@
 
 #include "runtime_layer_data.h"
 
-#include <inttypes.h>
+#include <cstdint>
+#include <iomanip>
+#include <vector>
 
 #include "debug_logging.h"
+#include "event_logging.h"
+#include "layer_utils.h"
 
 namespace performancelayers {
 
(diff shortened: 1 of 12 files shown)


layer/runtime_layer_data.cc
```diff
@@ -131,9 +135,20 @@ void RuntimeLayerData::LogAndRemoveQueryPools() {
       // FIXME: We should adjust the elapsed units to account for the current
       // GPU frequency. Calling vkGetPhysicalDeviceProperties here causes the
       // driver to crash, however.
-      Log("pipeline_execution", GetPipelineHash(info->pipeline),
-          absl::StrCat(timestamp1 - timestamp0, ",", invocations[0], ",",
-                       invocations[1]));
+      HashVector pipeline = GetPipelineHash(info->pipeline);
+      std::vector<int64_t> hashes(pipeline.begin(), pipeline.end());
+      RuntimeEvent event("pipeline_execution", hashes,
+                         DurationClock::duration(timestamp1 - timestamp0),
```
kuhar: Does this assume nanosecond-level precision? Would it be possible to call a function that explicitly constructs duration with some unit?

layer/runtime_layer_data.cc
```diff
@@ -131,9 +135,20 @@ void RuntimeLayerData::LogAndRemoveQueryPools() {
...
+                         invocations[0], invocations[1]);
+      LogEvent(&event);
+
+      std::stringstream pipeline_hash_str;
+      pipeline_hash_str << std::quoted(PipelineHashToString(pipeline));
```
kuhar, line 146: Hmm, since this pattern keeps coming up, maybe we should bring back the QuoteStr function and implement it in terms of `std::quoted`?

This can be done in a separate PR, no need to change this line

layer/runtime_layer_data.cc
```diff
@@ -131,9 +135,20 @@ void RuntimeLayerData::LogAndRemoveQueryPools() {
...
+      std::string pipeline_and_content =
+          CsvCat(pipeline_hash_str.str(),
+                 absl::StrCat(timestamp1 - timestamp0, ",", invocations[0], ",",
```
kuhar: Isn't this the same as a nested CsvCat?

kuhar (changes requested):

layer/runtime_layer_data.cc
miladHakimi: This code will be removed in the future PRs because the logger will take care of the quoted strings.

layer/runtime_layer_data.cc
miladHakimi: Yes, it assumes the nanosecond-level precision. We can have a function like `CreateDurationInNs()` that accepts 2 `int`s as the start and end of a duration and returns their differences. In the comments we can say the inputs must be in nanoseconds precision. Does it make sense?

kuhar: Yes, that sounds great!

kuhar: Or you could add your own class that wraps `DurationClock::duration` and has a static function to construct from nanoseconds, e.g.:
```c++
class Duration {
public:
  static Duration FromNanoseconds(int64_t nanos) { ... };
  int64_t ToNanoseconds() const { ... };

private:
  DurationClock::duration duration_;
};
```

miladHakimi: This looks cleaner.

> Or you could add your own class that wraps `DurationClock::duration` and has a static function to construct from nanoseconds, e.g.:
> 
> ```c++
> class Duration {
> public:
>   static Duration FromNanoseconds(int64_t nanos) { ... };
> 
> private:
>   DurationClock::duration duration_;
> };
> ```

This looks cleaner. Thanks!

kuhar (approved):
awesome

Issue #107, opened 2022-11-04
Update the CI for the common log file.


miladHakimi:
The format of the events written to the common log file has changed. An example of the new format:
`create_graphics_pipelines,timestamp:1667521492863138945,hashes:"[0x57dad0c231e4e5ce,0x92cc37a4b4360ca9]",duration:595730`
The `CommonLogger` logs the events in the new format while the current CI checks the common log file with the previous format. 
When we start refactoring the layers with `CommonLogger`, common log checks should be disabled and re-enabled when the checks are consistent with the new format.

miladHakimi:
In order not to break things while adding the `CommonLogger` to the layers, it's decided to incrementally introduce the new format to the `FileCheck` as we add the `CommonLogger` to each layer. See the discussion in #108

miladHakimi:
This issue is addressed by #108, #111, and #113.

PR #108, opened 2022-11-04
Add `CommonLogger` to the `LayerDataWithLogger` class.

miladHakimi:
This PR:
- Implements a new class, `LayerDataWithCommonLogger`, that contains the `CommonLogger`.
- Adds the new class to the memory usage layer.
- Change the format of the `FileCheck` input for the memory usage shared events.. Issue: #107

### DIFF
--- a/layer/common_logging.cc
+++ b/layer/common_logging.cc
@@ -27,49 +27,46 @@ std::string EventToCommonLogStr(Event &event) {
   std::ostringstream csv_str;
   csv_str << event.GetEventName() << "," << event.GetCreationTime().GetName()
           << ":" << ValueToCSVString(event.GetCreationTime().GetValue());
-  csv_str << ",";
-  for (size_t i = 0, e = attributes.size(); i != e; ++i) {
-    csv_str << attributes[i]->GetName() << ":";
-    switch (attributes[i]->GetValueType()) {
+  for (Attribute *attribute : attributes) {
+    csv_str << "," << attribute->GetName() << ":";
+    switch (attribute->GetValueType()) {
       case ValueType::kTimestamp: {
         csv_str << ValueToCSVString(
-            attributes[i]->cast<TimestampAttr>()->GetValue());
+            attribute->cast<TimestampAttr>()->GetValue());
         break;
       }
       case ValueType::kDuration: {
         csv_str << ValueToCSVString(
-            attributes[i]->cast<DurationAttr>()->GetValue());
+            attribute->cast<DurationAttr>()->GetValue());
         break;
       }
       case ValueType::kBool: {
-        csv_str << ValueToCSVString(
-            attributes[i]->cast<BoolAttr>()->GetValue());
+        csv_str << ValueToCSVString(attribute->cast<BoolAttr>()->GetValue());
         break;
       }
       case ValueType::kInt64: {
-        csv_str << ValueToCSVString(
-            attributes[i]->cast<Int64Attr>()->GetValue());
+        csv_str << ValueToCSVString(attribute->cast<Int64Attr>()->GetValue());
         break;
       }
       case ValueType::kString: {
-        csv_str << ValueToCSVString(
-            attributes[i]->cast<StringAttr>()->GetValue());
+        csv_str << ValueToCSVString(attribute->cast<StringAttr>()->GetValue());
         break;
       }
       case ValueType::kVectorInt64: {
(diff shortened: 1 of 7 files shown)


kuhar:
> Disables FileCheck for the common log file because it is checking based on the previous format. Issue: https://github.com/google/vulkan-performance-layers/issues/107

We should update the code and the tests at the same time. This will make code review easier and also prevent accidental breakage in between the PR that updates tests

miladHakimi:
> > Disables FileCheck for the common log file because it is checking based on the previous format. Issue: #107
> 
> We should update the code and the tests at the same time. This will make code review easier and also prevent accidental breakage in between the PR that updates tests



> > Disables FileCheck for the common log file because it is checking based on the previous format. Issue: #107
> 
> We should update the code and the tests at the same time. This will make code review easier and also prevent accidental breakage in between the PR that updates tests

In that case, we need one big PR that adds the `CommonLogger` to all the layers and fixes the `FileCheck`. I can do it in multiple commits to make it easier to follow. Does that make sense?

kuhar:
> In that case, we need one big PR that adds the CommonLogger to all the layers and fixes the FileCheck. I can do it in multiple commits to make it easier to follow. Does that make sense?

I don't see how this would be different from the CSV logging work. You should be able to update one layer at a time + the relevant part of the common log test.

miladHakimi:
> > In that case, we need one big PR that adds the CommonLogger to all the layers and fixes the FileCheck. I can do it in multiple commits to make it easier to follow. Does that make sense?
> 
> I don't see how this would be different from the CSV logging work. You should be able to update one layer at a time + the relevant part of the common log test.

I wanted to add the `CommonLogger` to the `LayerDataWithLogger`. If I do this, all the layers will be affected because all of them are currently using `LayerDataWithLogger`. I can also add another class, `LayerDataWithBroadcastLogger` (or something similar) and incrementally change the layers such that they use the new class instead. When I'm finished with all the layers, I can rename it.

kuhar:
Oh, I see. So once we are done with CSV logging, we can move it to the base `LayerData` and get rid of `LayerDataWithLogger`. And for common logs, we can add a new derived class `LayerDataWithCommonLogger` that layers opt into one by one. WDYT?

miladHakimi:
> Oh, I see. So once we are done with CSV logging, we can move it to the base `LayerData` and get rid of `LayerDataWithLogger`. And for common logs, we can add a new derived class `LayerDataWithCommonLogger` that layers opt into one by one. WDYT?

Makes sense. So, after we're done with the runtime layer we can do this refactor because the cache sideload layer doesn't have any private logs.

layer/common_logging.cc
```diff
@@ -27,9 +27,8 @@ std::string EventToCommonLogStr(Event &event) {
   std::ostringstream csv_str;
   csv_str << event.GetEventName() << "," << event.GetCreationTime().GetName()
           << ":" << ValueToCSVString(event.GetCreationTime().GetValue());
-  csv_str << ",";
   for (size_t i = 0, e = attributes.size(); i != e; ++i) {
```
kuhar: Can we use a range-based for loop now that we don't need to check for `i + 1`?

layer/common_logging.cc
```diff
@@ -62,14 +61,13 @@ std::string EventToCommonLogStr(Event &event) {
...
     }
-    if (i + 1 != e) csv_str << ",";
   }
   return csv_str.str();
 }
 
 CommonLogger::CommonLogger(const char *filename) {
   if (filename) {
-    out_ = fopen(filename, "w");
+    out_ = fopen(filename, "a");
```
kuhar, line 69: Can you add a short comment explaining why we want to append? (IUUC, because multiple layers will write to the same file at the same time)

layer/layer_data.cc
```diff
@@ -84,8 +84,8 @@ LayerData::LayerData(char* log_filename, const char* header)
   if (const char* event_log_file = getenv(kEventLogFileEnvVar)) {
     // The underlying log file can be written to by multiple layers from
     // multiple threads. All contentens have to be written in whole lines(s)
-    // at a time to ensure there is no unintended interleaving within a single
-    // line.
+    // at a time to ensure there is no unintended interleaving within a
+    // single line.
```
kuhar: Accidental change? Please undo if not required by clang-format.

kuhar (approved):
LGTM

PR #111, opened 2022-11-07
Refactor frametime layer to use `LayerDataWithCommonLogger`.

miladHakimi:
This PR:
- refactors frametime layer to use `EventLogger` for the common logs by making it use `LayerDataWithCommonLogger` instead of `LayerData`.
- Changes the `FileCheck` format for frametime layer.

Issues: #107 #112

### DIFF
--- a/layer/frame_time_layer.cc
+++ b/layer/frame_time_layer.cc
@@ -63,15 +63,37 @@ class FrameTimeEvent : public Event {
   BoolAttr started_;
 };
 
-class FrameTimeLayerData : public LayerData {
+// Frametime exit event. Both `application_exit` and `terminated` events can be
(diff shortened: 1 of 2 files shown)


kuhar:
Does this break the analysis scripts? If yes, I think that's fine given that there are no proper tests right now, but we should have at least a tracking issue added to the potentially broken scripts

miladHakimi:
> Does this break the analysis scripts? If yes, I think that's fine given that there are no proper tests right now, but we should have at least a tracking issue added to the potentially broken scripts

It breaks `plot_timeline.py` because timestamps are written in two different formats right now.

kuhar (approved):
LGTM

Issue #112, opened 2022-11-07
Broken Script Tracking Issue


miladHakimi:
### Broken Script
Changing the common log format causes the `plot_timeline.py`  script to break because it's using the previous format to parse the `event.log`. Currently, we're in the middle of refactoring the layers to use the `CommonLogger` that writes the events in the new format. We want to finish refactoring all the layers before updating the script. This issue is created to track the broken script and fix it after the refactoring is over.

We can fix this issue when the following layers are refactored:
- - [x] Frame time layer
- - [x] Compile time layer
- - [x] Runtime layer
- - [x] Cache sideload layer

- [x] Update `plot_timeline.py`

miladHakimi:
Now that the log format is consistent across the layers, we can update the script.
- [ ] Update `plot_timeline.py`

PR #113, opened 2022-11-07
Refactor compile time layer to use `CommonLogger`.

miladHakimi:
Introduce a new attribute, `HashAttr`, to differentiate between integers and shader hashes when logging.

Update `FileCheck` to use the new format for compile time common logs.

Issues: #107  #112


kuhar (approved):
LGTM

PR #117, opened 2022-11-08
Add `LogOutput` class.

### LINKED ISSUE #116: Introduce `LogOutput` abstraction.
After an `EventLogger` converts an `Event` to a string, it should write it to the output (Issue: #69 ). The output could be a file, standard output, or a string (for testing). `LogOutput` is the object that takes care of writing to the output and exposes the `LogLine()` method. Since the underlying file might be shared among multiple threads, `LogLine()` must be synchronized to make sure the logs do not get corrupted.
A sample implementation:
```c++
class Logoutput {
 public:
  virtual void LogLine(const std::string &line) = 0;
};

class FileOutput : public LogOutput {
 public:
  void LogLine(const std::string &line) override { ... }
 private:
  FILE *out_ = nullptr;
};

class StringOutput: public LogOutput {
 public:
  void LogLine(const std::string &line) override { ... }
  
  const std::string &GetLog() { return out_; } 
 private:
  std::string out_ = "";
};
```

miladHakimi:
This class provides an abstraction over the output. All the subclasses must implement `LogLine()` method which receives a string and writes it to the output.

This PR also contains 2 implementations of `LogOutput`class. `FileOutput` handles writing into a file and standard output. `StringOutput` writes to a string and is used for testing.

Fixes: #116

### DIFF
--- a/units/log_output_tests.cc
+++ b/units/log_output_tests.cc
@@ -1,0 +1,180 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include <array>
+#include <cassert>
+#include <cstddef>
+#include <cstdio>
+#include <fstream>
+#include <iostream>
+#include <string>
+#include <string_view>
+#include <thread>
+#include <vector>
+
+#include "debug_logging.h"
+#include "gmock/gmock.h"
+#include "gtest/gtest-spi.h"
+#include "gtest/gtest.h"
+
+using ::testing::ElementsAre;
+using ::testing::TempDir;
+
+namespace performancelayers {
+namespace {
+constexpr unsigned kLogsPerThread = 1337;
+constexpr unsigned kThreadCount = 100;
+
+// An abstraction over the output. It writes the incoming string to the output.
+class LogOutput {
+ public:
+  virtual ~LogOutput() = default;
+
+  virtual void Flush() = 0;
+
+  virtual void LogLine(std::string_view line) = 0;
+};
+
+// Implements LogOutput for a file. If the given filename is `nullptr`, it
+// writes to the standard output. After each write, the logs are persisted to
+// the file by calling `fflush`. Since only one line is written to the output by
+// each `LogLine()` call, it doens't need to aquire a mutex.
+class FileOutput : public LogOutput {
+ public:
+  FileOutput(const char *filename) {
+    if (!filename) {
+      out_ = stderr;
+      return;
+    }
+
+    // Since multiple layers open the same file and write to it at the same
+    // time, it's opened in the append mode.
+    out_ = fopen(filename, "a");
+    if (!out_) {
+      SPL_LOG(ERROR) << "Failed to open " << filename
+                     << ". Using stderr as the alternative output.";
+      out_ = stderr;
+    }
+  }
+
+  ~FileOutput() {
+    if (out_ && out_ != stderr) {
+      fclose(out_);
+    }
+  }
+
+  void Flush() override {
+    assert(out_);
+    fflush(out_);
+  }
+
+  void LogLine(std::string_view line) override {
+    assert(out_);
+    assert(line.find('\n') == std::string_view::npos &&
+           "Expected single line.");
+    const int line_len = line.length();
+    fprintf(out_, "%.*s\n", line_len, line.data());
+    Flush();
+  }
+
+ private:
+  FILE *out_ = nullptr;
+};
+
+// This class is used for testing. It writes the data to a string
+// instead of a file. The data can be read using the `GetLog()` method.
+class StringOutput : public LogOutput {
+ public:
+  StringOutput(){};
+
+  void Flush() override {}
+
+  void LogLine(std::string_view line) override {
+    assert(line.find('\n') == std::string_view::npos &&
+           "Expected single line.");
+    out_.push_back(line);
+  }
+
+  const std::vector<std::string_view> &GetLog() { return out_; }
+
+ private:
+  std::vector<std::string_view> out_;
+};
+
+// Used by the tests prior to logging to make sure the underlying file is empty.
+void CreateEmptyFile(const std::string &filename) {
+  std::ofstream(filename.c_str());
+}
+
+TEST(LogOutput, SingleThreadedLog) {
+  std::string file_path = TempDir() + "/single_thread.log";
+  CreateEmptyFile(file_path);
+
+  const std::string line = "name:event_name,timestamp:1234";
+  FileOutput file_out(file_path.c_str());
+  file_out.LogLine(line);
+
+  std::ifstream test_file(file_path);
+  ASSERT_TRUE(test_file.is_open());
+  std::string stored_line;
+  getline(test_file, stored_line);
+  EXPECT_EQ(stored_line, line);
+
+  StringOutput string_out;
+  string_out.LogLine(line);
+  EXPECT_THAT(string_out.GetLog(), ElementsAre(line));
+}
+
+#ifndef NDEBUG
+TEST(LogOutput, InvalidLog) {
+  std::string file_path = TempDir() + "/invalid.log";
+  CreateEmptyFile(file_path);
+  FileOutput file_out(file_path.c_str());
+  const std::string invalid_line =
+      "name:event_name,timestamp:1234\nname:event_name,timestamp:1234";
+  ASSERT_DEATH(file_out.LogLine(invalid_line), "Expected single line.");
+}
+#endif
+
+// This function is executed by multiple threads. It logs the input to the file
+// as many times as `kLogsPerThread`.
+void LogToFileOutput(FileOutput *file_out, std::string_view line) {
+  for (size_t i = 0; i < kLogsPerThread; ++i) file_out->LogLine(line);
+}
+
+TEST(LogOutput, MultiThreadedFileLog) {
+  std::string file_path = TempDir() + "/multi_thread.log";
+  CreateEmptyFile(file_path);
+
+  FileOutput file_out(file_path.c_str());
+  const std::string line = "name:event_name,timestamp:1234";
+  std::array<std::thread, kThreadCount> threads;
+  for (std::thread &thread : threads)
+    thread = std::thread(LogToFileOutput, &file_out, line);
+
+  for (std::thread &thread : threads) thread.join();
+
+  std::ifstream test_file(file_path);
+  ASSERT_TRUE(test_file.is_open());
+  std::string stored_line;
+  for (size_t i = 0; i < kThreadCount * kLogsPerThread; ++i) {
+    getline(test_file, stored_line);
+    EXPECT_EQ(stored_line, line);
+  }
+  getline(test_file, stored_line);
+  EXPECT_EQ(stored_line, "");
+}
+
+}  // namespace
+}  // namespace performancelayers
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -146,6 +146,7 @@ add_executable(layer_support_tests
     units/csv_log_tests.cc
     units/event_log_tests.cc
     units/input_buffer_tests.cc
+    units/log_output_tests.cc
     units/log_scanner_tests.cc
 )
 target_include_directories(layer_support_tests PRIVATE


units/log_output_tests.cc
```diff
@@ -0,0 +1,141 @@
...
+#include <cstdio>
+#include <fstream>
+#include <string>
+#include <thread>
+
+#include "debug_logging.h"
+#include "gtest/gtest.h"
+
+#define LOGS_PER_THREAD 1337
+#define THREAD_COUNT 100
```
kuhar: use constants instead, e.g.: `constexpr unsigned kLogsPerThread = 1337` inside the inner namespace

units/log_output_tests.cc
```diff
@@ -0,0 +1,141 @@
...
+
+namespace performancelayers {
+namespace {
+// An abstraction over the output. It writes the incoming string to the output.
+class LogOutput {
+ public:
+  virtual ~LogOutput() = default;
+
+  virtual void LogLine(const std::string &line) = 0;
```
kuhar: What about `Flush`? We will need it eventually.

units/log_output_tests.cc
```diff
@@ -0,0 +1,141 @@
...
+};
+
+// Implements LogOutput for a file. If the given filename is `nullptr`, it
+// writes to the standard output. After each write, the logs are persisted to
+// the file by calling `fflush`. Since only one line is written to the output by
+// each `LogLine()` call, it doens't need to aquire a mutex.
+class FileOutput : public LogOutput {
+ public:
+  FileOutput(const char *filename) {
+    if (filename) {
```
kuhar: nit: prefer early exits, e.g.:
```c++
if (!filename) {
  out_ = stderr;
  return;
}

out_ = fopen(...);
```

See https://llvm.org/docs/CodingStandards.html#use-early-exits-and-continue-to-simplify-code. There's also a similar rule made specifically for the google c++ coding style guide but I can't find it at the moment (should be one of https://abseil.io/tips/)

units/log_output_tests.cc
```diff
@@ -0,0 +1,141 @@
...
+      }
+    } else {
+      out_ = stderr;
+    }
+  }
+
+  ~FileOutput() {
+    if (out_ && out_ != stderr) {
+      fclose(out_);
+      out_ = nullptr;
```
kuhar: nit: we can't run any other code after the destructor so we don't have to zero-out values. Feel free to keep it if you found it useful when debugging etc

units/log_output_tests.cc
```diff
@@ -0,0 +1,141 @@
...
+class StringOutput : public LogOutput {
+ public:
+  StringOutput(){};
+
+  void LogLine(const std::string &line) override { out_ += line + "\n"; }
+
+  const std::string &GetLog() { return out_; }
+
+ private:
+  std::string out_ = "";
```
kuhar: What do you think about keeping a vector or list of lines instead? This should be easier to check with `ElementsAre` and `ElementsAreArray` matchers

units/log_output_tests.cc
```diff
@@ -0,0 +1,141 @@
...
+    EXPECT_EQ(stored_line, line);
+  }
+  getline(test_file, stored_line);
+  EXPECT_EQ(stored_line, "");
+  test_file.close();
+  remove(filename);
+}
+
+}  // namespace
+}  // namespace performancelayers
```
kuhar: nit: missing newline at the end

kuhar (changes requested):

units/log_output_tests.cc
miladHakimi: Thanks! It was really helpful!

units/log_output_tests.cc
miladHakimi: Yeah. It's cleaner this way.

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+  StringOutput(){};
+
+  void Flush() override {}
+
+  void LogLine(const std::string &line) override { out_.push_back(line); }
+
+  const std::vector<std::string> &GetLog() { return out_; }
+
+ private:
+  std::vector<std::string> out_ = {};
```
kuhar: nit: remove the initialization `= {};`. We require initialization for primitive types custom structs/classes, but there's no need to initialize members of types with well-known default constructors like `vector`, `string`, and other containers in the standard library.

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+constexpr unsigned kThreadCount = 100;
+
+// An abstraction over the output. It writes the incoming string to the output.
+class LogOutput {
+ public:
+  virtual ~LogOutput() = default;
+
+  virtual void Flush() = 0;
+
+  virtual void LogLine(const std::string &line) = 0;
```
kuhar: nit: I think it would be better to take `std::string_view` as the argument so that we don't force strings to be null-terminated and avoid allocations for string literals. I don't think it matters much though, so feel free to keep this as-is.

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+
+// This class is used for testing. It writes the data to a string
+// instead of a file. The data can be read using the `GetLog()` method.
+class StringOutput : public LogOutput {
+ public:
+  StringOutput(){};
+
+  void Flush() override {}
+
+  void LogLine(const std::string &line) override { out_.push_back(line); }
```
kuhar: Should we add an assertion that checks that logged lines don't have unexpected newline characters? I don't think we should support logging more than one line at the time though this function.

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+  const char *filename = "test.log";
+  FileOutput file_out(filename);
+  const std::string line = "name:event_name,timestamp:1234";
+  std::thread threads[kThreadCount];
+
+  for (size_t i = 0; i < kThreadCount; ++i)
+    threads[i] = std::thread(LogToFileOutput, &file_out, line);
+
+  for (size_t i = 0; i < kThreadCount; ++i) threads[i].join();
+  ;
```
kuhar: we don't need this one

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+  EXPECT_EQ(stored_line, line);
+  test_file.close();
+  remove(filename);
+
+  StringOutput string_out;
+  string_out.LogLine(line);
+  EXPECT_THAT(string_out.GetLog(), ElementsAre(line));
+}
+
+// This functions is executed by multiple threads. It logs the input to the file
```
kuhar: s/functions/function/

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+  const std::vector<std::string> &GetLog() { return out_; }
+
+ private:
+  std::vector<std::string> out_ = {};
+};
+
+TEST(LogOutput, SingleThreadedLog) {
+  const char *filename = "test.log";
+  const std::string line = "name:event_name,timestamp:1234";
+  FileOutput file_out(filename);
```
kuhar: I think this will create the test file in the current working directory. Shouldn't we write it somewhere under `/tmp`?
There should be a helper function for this. I found these these two:
1. https://github.com/google/googletest/pull/1077/files
2. https://en.cppreference.com/w/cpp/filesystem/temp_directory_path

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+
+// This functions is executed by multiple threads. It logs the input to the file
+// as many times as `kLogsPerThread`.
+void LogToFileOutput(FileOutput *file_out, const std::string &line) {
+  for (size_t i = 0; i < kLogsPerThread; ++i) file_out->LogLine(line);
+}
+
+TEST(LogOutput, MultiThreadedFileLog) {
+  const char *filename = "test.log";
+  FileOutput file_out(filename);
```
kuhar: also here

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+  const std::string line = "name:event_name,timestamp:1234";
+  std::thread threads[kThreadCount];
```
kuhar: use std::array instead

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+
+  for (size_t i = 0; i < kThreadCount; ++i)
+    threads[i] = std::thread(LogToFileOutput, &file_out, line);
+
+  for (size_t i = 0; i < kThreadCount; ++i) threads[i].join();
```
kuhar: use range-for

units/log_output_tests.cc
```diff
@@ -0,0 +1,156 @@
...
+  EXPECT_TRUE(test_file.is_open());
+  std::string stored_line;
+  for (size_t i = 0; i < kThreadCount * kLogsPerThread; ++i) {
+    getline(test_file, stored_line);
+    EXPECT_EQ(stored_line, line);
+  }
+  getline(test_file, stored_line);
+  EXPECT_EQ(stored_line, "");
+  test_file.close();
+  remove(filename);
```
kuhar: If we create this in a temp dir we should not have to remove the test files, so that someone can read the log file contents when they have to debug it

kuhar (changes requested):

units/log_output_tests.cc
miladHakimi: Since we append to the event log file, the previously written logs will be there the second time we run the checks. We can clear the files at the beginning of the tests.

units/log_output_tests.cc
```diff
@@ -0,0 +1,167 @@
...
+  test_file.open(file_path);
+  EXPECT_TRUE(test_file.is_open());
+  std::string stored_line;
+  for (size_t i = 0; i < kThreadCount * kLogsPerThread; ++i) {
+    getline(test_file, stored_line);
+    EXPECT_EQ(stored_line, line);
+  }
+  getline(test_file, stored_line);
+  EXPECT_EQ(stored_line, "");
+  test_file.close();
```
kuhar: no need to close manually, the destructor will handle this: https://en.cppreference.com/w/cpp/io/basic_fstream#:~:text=(destructor),virtual%5D%20(implicitly%20declared)

units/log_output_tests.cc
```diff
@@ -0,0 +1,167 @@
...
+  const std::string line = "name:event_name,timestamp:1234";
+  std::array<std::thread, kThreadCount> threads;
+  for (std::thread &thread : threads)
+    thread = std::thread(LogToFileOutput, &file_out, line);
+
+  for (std::thread &thread : threads) thread.join();
+
+  std::ifstream test_file;
+  test_file.open(file_path);
+  EXPECT_TRUE(test_file.is_open());
```
kuhar: we should abort here if the file is not open, use `ASSERT_TRUE` instead

units/log_output_tests.cc
```diff
@@ -0,0 +1,167 @@
...
+  std::string file_path = TempDir() + "/single_thread.log";
+  CreateEmptyFile(file_path);
+
+  const std::string line = "name:event_name,timestamp:1234";
+  FileOutput file_out(file_path.c_str());
+  file_out.LogLine(line);
+
+  std::ifstream test_file;
+  test_file.open(file_path);
+  EXPECT_TRUE(test_file.is_open());
```
kuhar: ASSERT_TRUE

units/log_output_tests.cc
```diff
@@ -0,0 +1,167 @@
...
+      fclose(out_);
+    }
+  }
+
+  void Flush() override {
+    assert(out_);
+    fflush(out_);
+  }
+
+  void LogLine(const std::string_view &line) override {
```
kuhar: `string_view` and similar 'view' types (e.g., `llvm::ArrayRef`, `llvm::StringRef`, `std::span`) are  cheap to copy, you can pass this by value

units/log_output_tests.cc
```diff
@@ -0,0 +1,167 @@
...
+
+// This class is used for testing. It writes the data to a string
+// instead of a file. The data can be read using the `GetLog()` method.
+class StringOutput : public LogOutput {
+ public:
+  StringOutput(){};
+
+  void Flush() override {}
+
+  void LogLine(const std::string_view &line) override {
```
kuhar: also here

units/log_output_tests.cc
```diff
@@ -0,0 +1,167 @@
...
+constexpr unsigned kThreadCount = 100;
+
+// An abstraction over the output. It writes the incoming string to the output.
+class LogOutput {
+ public:
+  virtual ~LogOutput() = default;
+
+  virtual void Flush() = 0;
+
+  virtual void LogLine(const std::string_view &line) = 0;
```
kuhar: also here

units/log_output_tests.cc
```diff
@@ -0,0 +1,167 @@
...
+
+  void Flush() override {
+    assert(out_);
+    fflush(out_);
+  }
+
+  void LogLine(const std::string_view &line) override {
+    assert(out_);
+    assert(!line.find('\n'));
+    fprintf(out_, "%s\n", line.data());
```
kuhar: line may not be null terminated now that we switched to `string_view`. We have to use the variable-length string format: https://man7.org/linux/man-pages/man3/printf.3.html#:~:text='*'%20(see%20field%20width%20and%20precision%20below)

kuhar (changes requested):

units/log_output_tests.cc
kuhar: We should also add a test for this case

units/log_output_tests.cc
```diff
@@ -0,0 +1,167 @@
...
+  }
+
+  void Flush() override {
+    assert(out_);
+    fflush(out_);
+  }
+
+  void LogLine(const std::string_view &line) override {
+    assert(out_);
+    assert(!line.find('\n'));
```
kuhar: Can you add an assertion message that explains why this is wrong? Something like `&& "Expected single line")`

units/log_output_tests.cc
kuhar: This doesn't seem done

units/log_output_tests.cc
```diff
@@ -0,0 +1,179 @@
...
+TEST(LogOutput, SingleThreadedLog) {
+  std::string file_path = TempDir() + "/single_thread.log";
+  CreateEmptyFile(file_path);
+
+  const std::string line = "name:event_name,timestamp:1234";
+  FileOutput file_out(file_path.c_str());
+  file_out.LogLine(line);
+
+  std::ifstream test_file;
+  test_file.open(file_path);
```
kuhar: nit: you can give the file path to the constructor to combine these two

units/log_output_tests.cc
miladHakimi: Sorry. Missed this one.

units/log_output_tests.cc
```diff
@@ -0,0 +1,179 @@
...
+
+// Used by the tests prior to logging to make sure the underlying file is empty.
+void CreateEmptyFile(std::string &filename) { std::ofstream(filename.c_str()); }
+
+TEST(LogOutput, SingleThreadedLog) {
+  std::string file_path = TempDir() + "/single_thread.log";
+  CreateEmptyFile(file_path);
+
+  const std::string line = "name:event_name,timestamp:1234";
+  FileOutput file_out(file_path.c_str());
```
kuhar, line 125: I don't think you need `c_str`? This would be slow because `string_view` would have to call `strlen` on the c-string instead of using the known length. Instead, either convert it directly to `string_view` or make `line` itself a `string_view`

units/log_output_tests.cc
```diff
@@ -0,0 +1,179 @@
...
+  StringOutput string_out;
+  string_out.LogLine(line);
+  EXPECT_THAT(string_out.GetLog(), ElementsAre(line));
+}
+
+#ifndef NDEBUG
+TEST(LogOutput, InvalidLog) {
+  std::string file_path = TempDir() + "/invalid.log";
+  CreateEmptyFile(file_path);
+  FileOutput file_out(file_path.c_str());
```
kuhar, line 143: also here

units/log_output_tests.cc
```diff
@@ -0,0 +1,179 @@
...
+  const std::string line = "name:event_name,timestamp:1234";
+  std::array<std::thread, kThreadCount> threads;
+  for (std::thread &thread : threads)
+    thread = std::thread(LogToFileOutput, &file_out, line);
+
+  for (std::thread &thread : threads) thread.join();
+
+  std::ifstream test_file;
+  test_file.open(file_path);
```
kuhar: also here

kuhar (changes requested):

units/log_output_tests.cc
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  const std::vector<std::string_view> &GetLog() { return out_; }
+
+ private:
+  std::vector<std::string_view> out_;
+};
+
+// Used by the tests prior to logging to make sure the underlying file is empty.
+void CreateEmptyFile(std::string &filename) { std::ofstream(filename.c_str()); }
```
kuhar: add const

units/log_output_tests.cc
kuhar: disregard, this creates a new file from the path

kuhar (approved):
LGTM, thanks for the fixes

PR #118, opened 2022-11-09
Update plot_timeline script to use the new log format.

### LINKED ISSUE #112: Broken Script Tracking Issue

miladHakimi:
Updating the event log format causes this script to break because it was using the old format. This PR fixes this issue.

Fixes: #112

### DIFF
--- a/scripts/plot_timeline.py
+++ b/scripts/plot_timeline.py
@@ -27,17 +27,61 @@ To see full command line options documentation, run:
   plot_timeline.py --help
 """
 
+import re
+import pandas as pd
+import sys
+import numpy as np
+import matplotlib.pyplot as plt
 import argparse
 import csv
 import matplotlib
-matplotlib.use('Agg') # Allows to work without X dispaly
-import matplotlib.pyplot as plt
-import numpy as np
-import sys
-import pandas as pd
+matplotlib.use('Agg')  # Allows to work without X dispaly
 
 nanos_per_second = 10**9
 nanos_per_milli = 10**6
+
+
+class FrameEvent:
+    """
+    A datastructure containing Frame event data
+
+    A frame event starts with `frame_present` and contains frame_time
+    and the frame state information.
+    """
+
+    def __init__(self, timestamp: int, event_str: str):
+        self.timestamp = timestamp
+
+        pattern = r',frame_time:(\d+),started:([0-1])'
+        attributes = re.match(pattern, event_str)
+        assert (attributes)
+        attributes = attributes.groups()
+        assert (len(attributes) >= 2)
+
+        self.frame_time = int(attributes[0])
+        self.started = int(attributes[1])
+
+
+class PipelineEvent:
+    """
+    A datastructure containing Pipline event data.
+
+    A Pipeline event starts with `create_graphics_pipelines` or
+    `create_compute_pipelines` and contains the shader hashes and pipline duration.
+    """
+
+    def __init__(self, timestamp: int, event_str: str):
+        self.timestamp = timestamp
+
+        pattern = r',hashes:"\[((0x[a-f0-9]*)+,*)+\]",duration:(\d+)'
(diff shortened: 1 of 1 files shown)


scripts/plot_timeline.py
```diff
@@ -1,11 +1,11 @@
-#!/usr/bin/env python3
+#!/ usr / bin / env python3
```
kuhar: this seems broken

scripts/plot_timeline.py
```diff
@@ -1,11 +1,11 @@
...
 # Copyright 2020 Google LLC
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
+# Licensed under the Apache License, Version 2.0(the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
 #
-#     https://www.apache.org/licenses/LICENSE-2.0
+# https:  // www.apache.org/licenses/LICENSE-2.0
```
kuhar: also here

scripts/plot_timeline.py
```diff
@@ -1,11 +1,11 @@
-#!/usr/bin/env python3
+#!/ usr / bin / env python3
 # Copyright 2020 Google LLC
 #
-# Licensed under the Apache License, Version 2.0 (the "License");
+# Licensed under the Apache License, Version 2.0(the "License");
```
kuhar: can you double check the license header?

scripts/plot_timeline.py
```diff
@@ -91,48 +137,86 @@ def get_pipeline_creation_times(create_pipeline_events):
...
+
+def parse_event_log(log: str) -> (str, int, str):
+    """
+    Returns a triple containing event name, timestamp, and rest of event which
+    may be empty or contain attributes.
+    log format sample:
+    sample1: `compile_time_layer_init,timestamp:123`
+    sample2: `frame_present,timestamp:1667942408738000395,frame_time:9707270,started:1`
+    """
+# This pattern captures `event_type`, `timestamp`, and the rest of the event.
```
kuhar: this comment is not aligned with the code

scripts/plot_timeline.py
```diff
@@ -91,48 +137,86 @@ def get_pipeline_creation_times(create_pipeline_events):
...
     ax.set_ylabel('Creation Time [ms]')
     max_creation_time_millis = 0
 
     if 'create_graphics_pipelines' in events_by_type:
-        xs, ys = get_pipeline_creation_times(events_by_type['create_graphics_pipelines'])
-        ax.scatter(xs, ys, s=3, color=plt.cm.tab10(8), label='Create Graphics Pipelines')
+        xs, ys = get_pipeline_creation_times(
+            events_by_type['create_graphics_pipelines'])
+        ax.scatter(xs, ys, s=3, color=plt.cm.tab10(
+            8), label='Create Graphics Pipelines')
```
kuhar: I think we should keep these unrelated formatting changes separate. Either undo them here and elsewhere or open a PR to format the code you are going to modify first.

kuhar: Also elsewhere in this file.

kuhar (changes requested):

scripts/plot_timeline.py
miladHakimi: Sure!

kuhar (approved):

PR #119, opened 2022-11-09
Refactor `LayerData` to include `CommonLogger`.

miladHakimi:
We created the `LayerDataWithCommonLogger` class to incrementally introduce the `CommonLogger` to all the layers and update their corresponding `FileCheck`.

Now that all the layers can log using `CommonLogger` and all the tests pass, it's safe to add it to the `LayerData` and remove the `LayerDataWithCommonLogger` class.

Issue: #69


kuhar (approved):
LGTM

PR #122, opened 2022-11-14
Add runtime layer `FileCheck` to the CI.

### LINKED ISSUE #121: Add runtime layer FileCheck.
Now that the runtime layer doesn't hang in the CI when running `vkcube`, we can have a `FileCheck` for the output produced by the runtime layer.

miladHakimi:
Now that `vkcube` runs without a problem when the runtime layer is enabled, we can add checks for the runtime layer's logs in the CI. This PR implements the tests for both private and common files.
Fixes: #121


kuhar (approved):
Awesome

PR #127, opened 2022-11-15
Refactor `CommonLogger` to use `LogOutput`.

miladHakimi:
The `CommonLogger` receives a `LogOutput` pointer in its constructor that determines the type of output it wants to write the logs into.


kuhar (approved):
LGTM

PR #129, opened 2022-11-16
Improve source directory structure

### LINKED ISSUE #126: Improve source code directory structure
Currently all source and header files are in the same directory. This is not glob-friendly because we keep multiple binaries (layer `.so`s) in the same directory.

Instead, we should separate out common source files and create a source directory for each binary target.

kuhar:
-  Move layers and support code to seaprate directories
-  Use full paths in include directives
-  Clean up cmake logic
-  Move layer dylib and manifest install files to separate directories
-  Fix CI test scripts to pick up the new directories

Fixes: https://github.com/google/vulkan-performance-layers/issues/126


vettoreldaniele (approved):
LGTM, nice refactor.

miladHakimi (approved):
Thanks for taking care of this!
LGTM.

Issue #130, opened 2022-11-18
Event logging implementation for the Chrome Trace Event format.


miladHakimi:
Here is the base Trace Event format:
```json
{
  "name": "myName",
  "cat": "category,list",
  "ph": "B",
  "ts": 12345,
  "pid": 123,
  "tid": 456,
  "args": {
    "someArg": 1,
    "anotherArg": {
      "value": "my value"
    }
  }
```
Not all the fields are required and we can skip `cat`, `pid`, `tid`, and `args`.  But, we might need additional fields based on the phase (`ph`) we specify. Since our main goal is to show duration, the phase should be `X` which indicates an event with duration  in Trace Event (a Complete event). This event requires an additional duration (`dur`) field. A simple Complete event would look like this:
```json
{ "name": "myName", "ph": "X", "ts": 12345, "dur": 123}
``` 
Current `Event` class does not have a "phase" and putting the phase as an attribute might mess things up because we have to change the other loggers to ignore it. I suggest the following code:
```c++
enum Phase { Complete };

class TraceEvent: public Event { 
 public:
  Phase GetPhase() { return phase_;};
  virtual Duration GetDuration() { ... }
 private:
  Phase phase_;
};

std::string TraceEventToStr(TraceEvent *event) {
    std::stringstream event_str;
    int64_t timestamp_ms = ToMilliseconds(event->GetTimestamp());
    event_str << "{" << std::quoted("name") << ":" << std::quoted(event->GetName()) << std::quoted("ts") << ":" << timestamp_ms << "}";
    if (event->GetPhase() == Phase::Complete)
        event_str << std::quoted("ph") << ":" << std::quoted("X") << "," << std::quoted("dur") << ":" << ToMilliseconds(event->GetDuration());

    event_str << "},";    
    return event_str.str();    
}

```
To get the duration, the events that are created as a `TraceEvent` have already a duration attribute. There are a couple of ways to get that attribute:
- Keep a pointer to the duration attribute in the `TraceEvent` as a private variable. This approach might require the least modification to the previous code.
- Move the duration attribute to the `Event` class so that all the `Event`s have duration. The default duration should be 0. This might not make sense especially for the events that do not have any duration (e.g., the events indicating the start of a layer).   We should also modify the loggers as well because the durations are moved out of the attribute vector and we can't print them by just iterating over the attributes.
[...]

PR #132, opened 2022-11-23
Add helper functions for converting Trace Event to string.

miladHakimi:
- Add support for converting `Instant` and `Complete` events to string.
- Detect cases such as unknown phases, non-given args, potential miss-spellings in args, etc.
- Test the helper functions.

Issue: #130

### DIFF
--- a/layer/support/event_logging.h
+++ b/layer/support/event_logging.h
@@ -15,8 +15,10 @@
 #ifndef STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_EVENT_LOGGING_H_
 #define STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_EVENT_LOGGING_H_
 
+#include <algorithm>
 #include <cstdint>
 #include <string>
+#include <string_view>
 #include <vector>
 
 #include "layer/support/layer_utils.h"
@@ -133,9 +135,21 @@ class TimestampAttr : public Attribute {
 using BoolAttr = AttributeImpl<bool, ValueType::kBool>;
 using HashAttr = AttributeImpl<int64_t, ValueType::kHashAttribute>;
 using Int64Attr = AttributeImpl<int64_t, ValueType::kInt64>;
-using StringAttr = AttributeImpl<std::string, ValueType::kString>;
 using VectorInt64Attr =
     AttributeImpl<std::vector<int64_t>, ValueType::kVectorInt64>;
+
+class StringAttr : public Attribute {
+ public:
+  static constexpr ValueType id_ = ValueType::kString;
+
+  StringAttr(const char *name, const std::string &value)
+      : Attribute{name, ValueType::kString}, value_(value) {}
+
+  std::string_view GetValue() const { return value_; }
+
+ private:
+  std::string value_;
+};
 
 // A compound event used by all Trace Event types. Each event type must add
 // its required attribute(s) to args. Since the events don't own many args,
@@ -169,6 +183,23 @@ class TraceEventAttr : public Attribute {
 
   const std::vector<Attribute *> &GetArgs() const { return args_; }
 
+  template <typename AttrType>
+  const AttrType *GetArg() const {
+    for (Attribute *arg : args_) {
+      if (auto *attr = arg->cast<AttrType>()) return attr;
+    }
+    return nullptr;
+  }
+
+  template <typename AttrType>
+  const AttrType *GetArg(std::string_view name) const {
+    for (Attribute *arg : args_) {
+      if (auto *attr = arg->cast<AttrType>())
+        if (attr->GetName() == name) return attr;
+    }
+    return nullptr;
+  }
+
  private:
   StringAttr category_;
   StringAttr phase_;
@@ -199,6 +230,14 @@ class Event {
   LogLevel GetLogLevel() const { return log_level_; }
 
   const TimestampAttr &GetCreationTime() { return creation_time_; }
+
+  template <typename AttrType>
+  const AttrType *GetAttribute() const {
+    for (Attribute *attribute : attributes_) {
+      if (auto *attr = attribute->cast<AttrType>()) return attr;
+    }
+    return nullptr;
+  }
 
  protected:
   void InitAttributes(std::initializer_list<Attribute *> attrs) {
--- a/layer/unittest/trace_event_log_tests.cc
+++ b/layer/unittest/trace_event_log_tests.cc
@@ -12,7 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include <cassert>
 #include <cstdint>
+#include <iomanip>
+#include <sstream>
 #include <string>
 #include <string_view>
 #include <vector>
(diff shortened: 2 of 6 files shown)


layer/unittest/trace_event_log_tests.cc
```diff
@@ -27,8 +31,167 @@ using ::testing::IsEmpty;
...
+
+std::string ValueToJsonString(int64_t value) { return std::to_string(value); }
+
+std::string ValueToJsonString(const std::vector<int64_t> &values) {
+  std::stringstream json_str;
+  json_str << "[";
+  size_t e = values.size();
+  for (size_t i = 0; i != e; ++i) {
+    const char *delimiter = i < e - 1 ? "," : "";
+    json_str << "\"0x" << std::hex << values[i] << "\"" << delimiter;
```
kuhar: Wouldn't it be enough to pass `std::hex` only once (before the loop)?

layer/unittest/trace_event_log_tests.cc
```diff
@@ -27,8 +31,167 @@ using ::testing::IsEmpty;
...
 namespace performancelayers {
 namespace {
-// An `InstantEvent`. Adds a `scope` variable to the args indicating event's
-// visibility in the trace viewer.
+std::string ValueToJsonString(bool value) { return value ? "true" : "false"; }
+
+std::string ValueToJsonString(int64_t value) { return std::to_string(value); }
+
+std::string ValueToJsonString(const std::vector<int64_t> &values) {
+  std::stringstream json_str;
```
kuhar: nit: you can use `ostringstream` to indicate you will be writing only

layer/unittest/trace_event_log_tests.cc
```diff
@@ -27,8 +31,167 @@ using ::testing::IsEmpty;
...
+std::string ValueToJsonString(TimestampClock::time_point value) {
+  return std::to_string(ToUnixMillis(value));
+}
+
+// Appends all given attributes, including the type-specific attribute, to the
+// stream.
+void TraceArgsToJsonString(const std::vector<Attribute *> &args,
+                           std::ostringstream &json_str) {
+  json_str << ", " << std::quoted("args") << " : { ";
+  for (size_t i = 0, e = args.size(); i < e; ++i) {
```
kuhar, line 66: nit: prefer range-based for loops

layer/unittest/trace_event_log_tests.cc
```diff
@@ -27,8 +31,167 @@ using ::testing::IsEmpty;
...
+// given stream.
+void CompleteEventToString(TimestampAttr timestamp,
+                           const TraceEventAttr *trace_event,
+                           std::ostringstream &json_stream) {
+  const std::vector<Attribute *> &args = trace_event->GetArgs();
+  assert(args.size() && "CompleteEvent's args must not be empty.");
+
+  auto it = std::find_if(args.begin(), args.end(), [](Attribute *attr) {
+    return attr->isa<DurationAttr>();
+  });
```
kuhar: Idea: what do you think about adding a helper function to `TraceEventAttr` to retrieve attributes based on types? Something like:
```c++
auto *duration = trace_event->GetArg<DurationAttr>();
assert(duration && "Duration not found");
```

We could also add a similar helper to `Event`.

layer/unittest/trace_event_log_tests.cc
```diff
@@ -27,8 +31,167 @@ using ::testing::IsEmpty;
...
+  auto it = std::find_if(args.begin(), args.end(), [](Attribute *attr) {
+    return attr->isa<StringAttr>() &&
+           attr->GetName() == std::string_view("scope");
+  });
+  assert(it != args.end() && "Scope not found in args.");
+
+  const std::string &scope = (*it)->cast<StringAttr>()->GetValue();
+  assert((scope == std::string_view("g") || scope == std::string_view("p") ||
+          scope == std::string_view("t")) &&
+         "Invalid scope. Scope must be \"g\", \"p\", or \"t\".");
```
kuhar: nit: Hmm, should we make `StringAttr::GetValue` return a `string_view` instead of `char *`? The underlying value can still be `const char*`

layer/unittest/trace_event_log_tests.cc
```diff
@@ -27,8 +31,167 @@ using ::testing::IsEmpty;
...
+      case ValueType::kTraceEvent:
+        break;
+    }
+  }
+  json_str << " }";
+}
+
+// Appends the start timestamp and the duration of a `TraceEventAttr` to the
+// given stream.
+void CompleteEventToString(TimestampAttr timestamp,
```
kuhar: This function name is a little bit misleading because `Complete` is a verb. Since this does not return a `string`, can we call it `AppendCompleteEvent` or `PrintCompleteEvent`?

layer/unittest/trace_event_log_tests.cc
```diff
@@ -113,9 +276,180 @@ TEST(TraceEvent, CompleteEventCreation) {
...
+               << " : " << std::quoted("i") << ", " << std::quoted("cat")
+               << " : " << std::quoted("compile_time") << ", "
+               << std::quoted("pid") << " : " << 123 << ", "
+               << std::quoted("tid") << " : " << 321 << ", "
+               << std::quoted("ts") << " : " << std::fixed
+               << ToUnixMillis(instant_event.GetCreationTime().GetValue())
+               << ", " << std::quoted("s") << " : " << std::quoted("g") << ", "
+               << std::quoted("args") << " : "
+               << "{ " << std::quoted("scope") << " : " << std::quoted("g")
+               << " } },";
```
kuhar: I find this difficult to parse, there are many moving parts and calls here. Could we convert this to a [raw string literal](https://learn.microsoft.com/en-us/cpp/cpp/string-and-character-literals-cpp?view=msvc-170#raw-string-literals-c11)? E.g.: `string_view expected_str = R"({ "name": ...},)";`

if that's too long, you can add it in several chunks with `<<`, but it would be better to keep it as close to the actual string as possible.

Because creation time is not fixed, we can set it to some fixed value before printing.

layer/unittest/trace_event_log_tests.cc
```diff
@@ -113,9 +276,180 @@ TEST(TraceEvent, CompleteEventCreation) {
...
+  const int64_t hash_val2 = 0x67d390249c2f20ce;
+  Duration duration = Duration::FromNanoseconds(100000);
+  CreateShaderCompleteEvent complete_event(
+      "compile_time", {hash_val1, hash_val2}, duration, 123, 321);
+
+  std::ostringstream expected_str;
+  double start_clock =
+      ToUnixMillis(complete_event.GetCreationTime().GetValue()) -
+      duration.ToMilliseconds();
+  expected_str << "{ " << std::quoted("name") << " : "
```
kuhar: Same here

layer/unittest/trace_event_log_tests.cc
```diff
@@ -113,9 +276,180 @@ TEST(TraceEvent, CompleteEventCreation) {
...
+
+#ifndef NDEBUG
+TEST(TraceEvent, InvalidEventType) {
+  Event event("event without TraceEventAttr");
+  EXPECT_DEATH(EventToTraceEventString(event),
+               "Could not find TraceEventAttr in the event.");
+
+  UnhandledTypeEvent unrecognized_event = {};
+  ASSERT_DEATH(EventToTraceEventString(unrecognized_event),
+               "Unrecognized phase.");
```
kuhar: Nice! But shouldn't these be two separate test cases? Does the second one ever get executed when the first one dies?

layer/unittest/trace_event_log_tests.cc
```diff
@@ -113,9 +276,180 @@ TEST(TraceEvent, CompleteEventCreation) {
...
+        scope_(attr_name, attr_value),
+        trace_event_("trace_event", category, "i", pid, tid, {&scope_}) {
+    InitAttributes({&trace_event_});
+  }
+
+  StringAttr scope_;
+  TraceEventAttr trace_event_;
+};
+
+#ifndef NDEBUG
```
kuhar: Should we move this up to start above `InvalidInstantEvent`?

layer/unittest/trace_event_log_tests.cc
```diff
@@ -113,9 +276,180 @@ TEST(TraceEvent, CompleteEventCreation) {
...
+        hash_values_("hashes", {}),
+        trace_event_("trace_event", "pipeline", "X", pid, tid, {}) {
+    InitAttributes({&hash_values_, &trace_event_});
+  }
+
+  VectorInt64Attr hash_values_;
+  TraceEventAttr trace_event_;
+};
+
+#ifndef NDEBUG
```
kuhar, line 363: Also here

layer/unittest/trace_event_log_tests.cc
```diff
@@ -113,9 +276,180 @@ TEST(TraceEvent, CompleteEventCreation) {
...
+TEST(TraceEvent, InvalidInstantEvent) {
+  InvalidInstantEvent no_arg_event = InvalidInstantEvent::EventWithoutArgs();
+  EXPECT_DEATH(EventToTraceEventString(no_arg_event),
+               "InstantEvent's args must not be empty.");
+
+  InvalidInstantEvent no_scope_event =
+      InvalidInstantEvent::EventWithoutScopeArg();
+  EXPECT_DEATH(EventToTraceEventString(no_scope_event),
```
kuhar: Also here

layer/unittest/trace_event_log_tests.cc
```diff
@@ -113,9 +276,180 @@ TEST(TraceEvent, CompleteEventCreation) {
...
+
+#ifndef NDEBUG
+TEST(TraceEvent, InvalidCompleteEvent) {
+  InvalidCompleteEvent no_arg_event = InvalidCompleteEvent::EventWithoutArgs();
+  EXPECT_DEATH(EventToTraceEventString(no_arg_event),
+               "CompleteEvent's args must not be empty.");
+
+  InvalidCompleteEvent no_scope_event =
+      InvalidCompleteEvent::EventWithoutDurationArg();
+  ASSERT_DEATH(EventToTraceEventString(no_scope_event),
```
kuhar: Also here

kuhar (changes requested):

layer/unittest/trace_event_log_tests.cc
miladHakimi: True. I think I should also add a `std::dec` at the end when I'm done writing. Right?

layer/unittest/trace_event_log_tests.cc
miladHakimi: I forgot to put the comma separator at the end. I needed the normal loop for that.

layer/unittest/trace_event_log_tests.cc
miladHakimi: That's a good idea. Maybe I can overload it for a `string`  input as well to scan based on the name. This would be useful for `scope`.

layer/unittest/trace_event_log_tests.cc
miladHakimi: Sure, I was a bit iffy about the name too. Is `AppendCompleteEventToStream` too long?

layer/unittest/trace_event_log_tests.cc
kuhar: I don't think you need to pass `dec` because the stringstream is discarded at the end

layer/unittest/trace_event_log_tests.cc
kuhar: I'd think it obvious it's the stream it appends to, but up to you

layer/unittest/trace_event_log_tests.cc
kuhar: 👍

layer/unittest/trace_event_log_tests.cc
miladHakimi: I see.

layer/unittest/trace_event_log_tests.cc
miladHakimi: I think it's fine. When I change the matching string of the `ASSERT_DEATH`, it gives me errors.

kuhar: 👍

layer/support/event_logging.h
```diff
@@ -133,10 +135,25 @@ class TimestampAttr : public Attribute {
...
+ public:
+  static constexpr ValueType id_ = ValueType::kString;
+
+  StringAttr(const char *name, const std::string &value)
+      : Attribute{name, ValueType::kString}, value_(value) {}
+
+  std::string_view GetValue() const { return value_; }
+
+ protected:
+  std::string_view GetValue() { return value_; }
```
kuhar: I don't think we need this one

layer/support/event_logging.h
```diff
@@ -169,6 +186,21 @@ class TraceEventAttr : public Attribute {
 
   const std::vector<Attribute *> &GetArgs() const { return args_; }
 
+  template <class T>
```
kuhar: nit: `typename T` to be consistent with the rest of the codebase

layer/support/event_logging.h
```diff
@@ -169,6 +186,21 @@ class TraceEventAttr : public Attribute {
...
+  const T *GetArg() const {
+    auto it = std::find_if(args_.begin(), args_.end(),
+                           [](Attribute *arg) { return arg->isa<T>(); });
+    return it != args_.end() ? (*it)->template cast<T>() : nullptr;
+  }
+
+  template <class T>
```
kuhar: also here

layer/support/event_logging.h
kuhar: also, we can be more descriptive and make it clear what kind of type is expected: `typename AttrType`

layer/support/event_logging.h
```diff
@@ -169,6 +186,21 @@ class TraceEventAttr : public Attribute {
 
   const std::vector<Attribute *> &GetArgs() const { return args_; }
 
+  template <class T>
+  const T *GetArg() const {
+    auto it = std::find_if(args_.begin(), args_.end(),
+                           [](Attribute *arg) { return arg->isa<T>(); });
+    return it != args_.end() ? (*it)->template cast<T>() : nullptr;
```
kuhar: nit: consider replacing this with a for loop, it should be a bit easier to parse here

layer/support/event_logging.h
```diff
@@ -169,6 +186,21 @@ class TraceEventAttr : public Attribute {
...
+  }
+
+  template <class T>
+  const T *GetArg(const char *name) const {
+    auto it = std::find_if(args_.begin(), args_.end(), [name](Attribute *arg) {
+      return arg->isa<T>() && arg->cast<T>()->GetName() == name;
+    });
+    return it != args_.end() ? (*it)->template cast<T>() : nullptr;
```
kuhar: also here

layer/support/event_logging.h
```diff
@@ -200,6 +232,13 @@ class Event {
 
   const TimestampAttr &GetCreationTime() { return creation_time_; }
 
+  template <class T>
+  const T *GetAttribute() const {
+    auto it = std::find_if(attributes_.begin(), attributes_.end(),
+                           [](Attribute *attr) { return attr->isa<T>(); });
+    return it != attributes_.end() ? (*it)->template cast<T>() : nullptr;
```
kuhar: also here

layer/unittest/trace_event_log_tests.cc
```diff
@@ -96,26 +319,80 @@ TEST(TraceEvent, CompleteEventCreation) {
...
-  EXPECT_EQ((*dur_it)->cast<DurationAttr>()->GetValue().ToNanoseconds(),
-            duration.ToNanoseconds());
+  const DurationAttr *dur_attr = trace_attr->GetArg<DurationAttr>();
+  ASSERT_TRUE(dur_attr);
+}
+
+TEST(TraceEvent, InstantEventToString) {
+  InstantEvent instant_event("compile_time_init", "compile_time", 123, 321);
+  std::ostringstream expected_str;
+  // std::string_view first_expected_str =
```
kuhar: remove this

kuhar (changes requested):

layer/support/event_logging.h
```diff
@@ -169,6 +186,21 @@ class TraceEventAttr : public Attribute {
...
 
+  template <class T>
+  const T *GetArg() const {
+    auto it = std::find_if(args_.begin(), args_.end(),
+                           [](Attribute *arg) { return arg->isa<T>(); });
+    return it != args_.end() ? (*it)->template cast<T>() : nullptr;
+  }
+
+  template <class T>
+  const T *GetArg(const char *name) const {
```
kuhar: We could also make it `string_view name`

layer/support/event_logging.h
```diff
@@ -169,6 +183,23 @@ class TraceEventAttr : public Attribute {
...
+    for (Attribute *arg : args_) {
+      if (arg->isa<AttrType>()) return arg->cast<AttrType>();
+    }
+    return nullptr;
+  }
+
+  template <typename AttrType>
+  const AttrType *GetArg(std::string_view name) const {
+    for (Attribute *arg : args_) {
+      if (arg->isa<AttrType>() && arg->cast<AttrType>()->GetName() == name)
```
kuhar: nit: you can do this with 2 fewer `id` checks:
```c++
if (auto *attr = arg->cast<AttrType>(); attr->GetName() == name)
  return attr;
```

layer/support/event_logging.h
```diff
@@ -200,6 +231,14 @@ class Event {
 
   const TimestampAttr &GetCreationTime() { return creation_time_; }
 
+  template <typename AttrType>
+  const AttrType *GetAttribute() const {
+    for (Attribute *attr : attributes_) {
+      if (attr->isa<AttrType>()) return attr->cast<AttrType>();
```
kuhar: also here

layer/support/event_logging.h
```diff
@@ -169,6 +183,23 @@ class TraceEventAttr : public Attribute {
 
   const std::vector<Attribute *> &GetArgs() const { return args_; }
 
+  template <typename AttrType>
+  const AttrType *GetArg() const {
+    for (Attribute *arg : args_) {
+      if (arg->isa<AttrType>()) return arg->cast<AttrType>();
```
kuhar: also here

kuhar (approved):
Awesome, LGTM

layer/support/event_logging.h
```diff
@@ -169,6 +183,23 @@ class TraceEventAttr : public Attribute {
...
+    for (Attribute *arg : args_) {
+      if (auto *attr = arg->cast<AttrType>()) return attr;
+    }
+    return nullptr;
+  }
+
+  template <typename AttrType>
+  const AttrType *GetArg(std::string_view name) const {
+    for (Attribute *arg : args_) {
+      if (auto *attr = arg->cast<AttrType>(); attr->GetName() == name)
```
kuhar: ah, this also needs a null check `attr && attr->GetName() == name)`
Sorry, I forgot about this earlier

kuhar (changes requested):

layer/support/event_logging.h
kuhar: or alternatively you can decompose this into 2 ifs:
```c++
if (auto *attr = arg->cast<AttrType>())
  if (attr->GetName() == name)
    return attr;
```

kuhar (approved):

PR #133, opened 2022-11-23
Add `TraceEventLogger` implementation.

miladHakimi:
Issue: #130

### DIFF
--- a/layer/unittest/trace_event_log_tests.cc
+++ b/layer/unittest/trace_event_log_tests.cc
@@ -24,12 +24,14 @@
 #include "gtest/gtest.h"
 #include "layer/support/event_logging.h"
 #include "layer/support/layer_utils.h"
+#include "layer/support/log_output.h"
 
 using ::testing::ElementsAre;
-using ::testing::IsEmpty;
 
 namespace performancelayers {
 namespace {
+constexpr int64_t kTestEventTimestamp = 1401;
+
 std::string ValueToJsonString(bool value) { return value ? "true" : "false"; }
 
 std::string ValueToJsonString(int64_t value) { return std::to_string(value); }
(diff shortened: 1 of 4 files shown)


layer/unittest/trace_event_log_tests.cc
```diff
@@ -294,6 +294,36 @@ class InvalidCompleteEvent : public Event {
...
+  void StartLog() override {
+    assert(out_);
+    out_->LogLine("[");
+  }
+
+  // The ']' at the end of the JSON array is optional in the Trace Event format.
+  // We exploit this to allow multiple layers write into the same file.
+  void EndLog() override {}
+
+  void Flush() override { out_->Flush(); }
```
kuhar, line 319: assert? or alternatively, assert only in the constructor

layer/unittest/trace_event_log_tests.cc
```diff
@@ -393,5 +423,62 @@ TEST(TraceEventDeathTest, InvalidCompleteEvent) {
...
+  logger.StartLog();
+  logger.AddEvent(&instant_event);
+  logger.Flush();
+  logger.EndLog();
+  // Checks double `EndLog` calls.
+  logger.EndLog();
+  std::ostringstream expected_str;
+  expected_str
+      << R"({ "name" : "compile_time_init", "ph" : "i", "cat" : "compile_time", "pid" : 123, "tid" : 321, "ts" : )"
+      << std::fixed << ToUnixMillis(instant_event.GetCreationTime().GetValue())
```
kuhar: Can we pass some fixed creation time when we create the event or just after? I think this would make these tests a little bit easier -- we could embed the exact time  in the string

layer/unittest/trace_event_log_tests.cc
```diff
@@ -393,5 +423,62 @@ TEST(TraceEventDeathTest, InvalidCompleteEvent) {
...
+      << R"(, "s" : "g", "args" : { "scope" : "g" } },)";
+
+  std::ostringstream complete_expected_str;
+  double start_timestamp =
+      ToUnixMillis(complete_event.GetCreationTime().GetValue()) -
+      duration.ToMilliseconds();
+  complete_expected_str
+      << R"({ "name" : "compile_time", "ph" : "X", "cat" : "pipeline", "pid" : 321, "tid" : 123, "ts" : )"
+      << std::fixed << start_timestamp << R"(, "dur" : )"
```
kuhar: same here

layer/unittest/trace_event_log_tests.cc
```diff
@@ -294,8 +294,37 @@ class InvalidCompleteEvent : public Event {
...
+  // We exploit this to allow multiple layers write into the same file.
+  void EndLog() override {}
+
+  void Flush() override { out_->Flush(); }
+
+ private:
+  LogOutput *out_ = nullptr;
+};
+
+constexpr int64_t event_timestamp = 1401;
```
kuhar: Can you move this towards the top of the file and call it something like `kTestEventTimestamp`?

kuhar (approved):
LGTM

PR #134, opened 2022-11-24
Implement the `Timestamp` class to ease the usage of `time_point`.

miladHakimi:
- Allows creating `time_point`s from int values.
- Keeps track of timestamps' time unit.


kuhar (approved):
Nice cleanup!

PR #135, opened 2022-11-24
Add `TraceEventLogger` to the compile time layer.

miladHakimi:
- Move `TraceEventLogger` and its helper functions to a new header and source files.
- Implement `LayerDataWithTraceEventLogger`, a wrapper that adds the new logger to `LayerData`.
- Add `TraceEventAttr` to all of the compile time layer's events.
- Update the env vars exported in the CI script.

### DIFF
--- a/layer/compile_time/compile_time_layer.cc
+++ b/layer/compile_time/compile_time_layer.cc
@@ -25,6 +25,7 @@
 #include "layer/support/event_logging.h"
 #include "layer/support/layer_data.h"
 #include "layer/support/layer_utils.h"
+#include "layer/support/trace_event_logging.h"
 
 namespace performancelayers {
 namespace {
@@ -37,6 +38,7 @@ constexpr char kLayerName[] = "VK_LAYER_STADIA_pipeline_compile_time";
 constexpr char kLayerDescription[] =
     "Stadia Pipeline Compile Time Measuring Layer";
 constexpr char kLogFilenameEnvVar[] = "VK_COMPILE_TIME_LOG";
+constexpr char kTraceEventCategory[] = "compile_time_layer";
 
 class CompileTimeEvent : public Event {
  public:
@@ -44,13 +46,16 @@ class CompileTimeEvent : public Event {
                    Duration duration)
       : Event(name, LogLevel::kHigh),
         hash_values_("hashes", hash_values),
-        duration_{"duration", duration} {
-    InitAttributes({&hash_values_, &duration_});
+        duration_{"duration", duration},
+        trace_attr_("trace_attr", kTraceEventCategory, "X",
+                    {&duration_, &hash_values_}) {
+    InitAttributes({&hash_values_, &duration_, &trace_attr_});
   }
 
  private:
   VectorInt64Attr hash_values_;
   DurationAttr duration_;
+  TraceEventAttr trace_attr_;
 };
 
 class ShaderModuleSlackEvent : public Event {
@@ -59,13 +64,16 @@ class ShaderModuleSlackEvent : public Event {
                          Duration duration)
       : Event(name),
         shader_hash_("shader_hash", hash_value),
-        duration_("slack", duration) {
-    InitAttributes({&shader_hash_, &duration_});
+        duration_("slack", duration),
+        trace_attr_("trace_attr", kTraceEventCategory, "X",
+                    {&duration_, &shader_hash_}) {
+    InitAttributes({&shader_hash_, &duration_, &trace_attr_});
   }
 
  private:
   HashAttr shader_hash_;
   DurationAttr duration_;
+  TraceEventAttr trace_attr_;
 };
 
 class CreateShaderEvent : public Event {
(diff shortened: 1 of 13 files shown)


layer/compile_time/compile_time_layer.cc
```diff
@@ -59,39 +63,60 @@ class ShaderModuleSlackEvent : public Event {
...
+    InitAttributes({&shader_hash_, &duration_, &trace_attr_});
   }
 
  private:
   HashAttr shader_hash_;
   DurationAttr duration_;
+  TraceEventAttr trace_attr_;
+};
+
+class CompileTimeStartEvent : public Event {
```
kuhar: I expected this event to denote compilation start but it seems like it's for layer initialization. Should we call it LayerInitEvent?

layer/support/layer_utils.cc
```diff
@@ -15,8 +15,30 @@
...
+}
+
+int64_t GetProcessId() {
+#ifdef __linux__
+  return getpid();
+#else
+  return 0;
+#endif
+}
 TimestampClock::time_point GetTimestamp() { return TimestampClock::now(); }
```
kuhar, line 45: Add a new line before the next function

layer/support/layer_utils.h
```diff
@@ -56,6 +56,17 @@
       reinterpret_cast<PFN_vk##FUNC_NAME_>(gdpa(*device, "vk" #FUNC_NAME_))
 
 namespace performancelayers {
+
+// Returns the thread id of the caller.
+// TODO:: This function works only on linux. We should add support for other
```
kuhar: nit TODO:: --> TODO:

layer/support/layer_utils.h
```diff
@@ -56,6 +56,17 @@
...
+// operating systems.
+int64_t GetThreadId();
+
+// Returns the process id of the caller.
+// TODO:: This function works only on linux. We should add support for other
```
kuhar: same here

layer/support/trace_event_logging.cc
```diff
@@ -0,0 +1,154 @@
...
+
+void TraceArgsToJsonString(const std::vector<Attribute *> &args,
+                           std::ostringstream &json_str) {
+  json_str << ", " << std::quoted("args") << " : { ";
+  for (size_t i = 0, e = args.size(); i < e; ++i) {
+    json_str << std::quoted(args[i]->GetName()) << " : ";
+    switch (args[i]->GetValueType()) {
+      case kHashAttribute: {
+        json_str << "\"0x" << std::hex << args[i]->cast<HashAttr>()->GetValue()
+                 << "\"" << std::hex;
```
kuhar: Do we need the second hex?

layer/support/layer_utils.cc
```diff
@@ -15,8 +15,30 @@
...
+#ifdef __linux__
+  return gettid();
+#else
+  return 0;
+#endif
+}
+
+int64_t GetProcessId() {
+#ifdef __linux__
+  return getpid();
```
kuhar: We can put this one in a static variable so that it gets cached.

layer/compile_time/compile_time_layer.cc
```diff
@@ -37,20 +38,23 @@ constexpr char kLayerName[] = "VK_LAYER_STADIA_pipeline_compile_time";
 constexpr char kLayerDescription[] =
     "Stadia Pipeline Compile Time Measuring Layer";
 constexpr char kLogFilenameEnvVar[] = "VK_COMPILE_TIME_LOG";
-
+constexpr char kTraceEventCategory[] = "compile_time_layer";
 class CompileTimeEvent : public Event {
```
kuhar, line 43: nit: Can you add an empty line before this class definition?

layer/support/trace_event_logging.h
```diff
@@ -0,0 +1,87 @@
...
+    if (!event->GetAttribute<TraceEventAttr>()) return;
+    std::string event_str = EventToTraceEventString(*event);
+    out_->LogLine(event_str);
+  }
+
+  // Writes '[', indicating the start point of the JSON array.
+  void StartLog() override { out_->LogLine("["); }
+
+  // The ']' at the end of the JSON array is optional in the Trace Event format.
+  // We exploit this to allow multiple layers write into the same file.
```
kuhar: It doesn't really do it, does it?

layer/compile_time/compile_time_layer.cc
kuhar: in the future, we could move it to a headers so that it can be shared by all layers if we find it useful

miladHakimi: > I expected this event to denote compilation start but it seems like it's for layer initialization. Should we call it LayerInitEvent?

Sure!

miladHakimi: > in the future, we could move it to a headers so that it can be shared by all layers if we find it useful

That's a good idea. I've implemented this as a `DefaultInstantEvent` and it'll be in the next PR.

layer/support/trace_event_logging.h
miladHakimi: We don't write ']' at the end of the log. I updated the description.

layer/support/layer_utils.cc
```diff
@@ -15,8 +15,33 @@
...
+  static thread_local int64_t tid = gettid();
+  return tid;
+#else
+  return 0;
+#endif
+}
+
+int64_t GetProcessId() {
+#ifdef __linux__
+  static thread_local int64_t pid = getpid();
```
kuhar: This doesn't have to be thead-local. All threads share the same PID.

layer/support/layer_utils.cc
```diff
@@ -15,8 +15,33 @@
...
+#ifdef __linux__
+#include <sys/types.h>
+#include <unistd.h>
+#endif
 
 namespace performancelayers {
+
+int64_t GetThreadId() {
+#ifdef __linux__
+  static thread_local int64_t tid = gettid();
```
kuhar, line 29: nice!

layer/support/layer_utils.cc
miladHakimi: But what if we have different processes? Will it be initialized in each process separately?

kuhar: Yes

kuhar: There's always only one process you have to worry about: the vulkan application process. If you want to coordinate across processes you need to use [some form of IPC](https://en.wikipedia.org/wiki/Inter-process_communication)

miladHakimi: Got it.

kuhar (approved):
LGTM

PR #136, opened 2022-11-25
Add `TraceEventLogger` to runtime layer.

miladHakimi:


layer/unittest/event_log_tests.cc
```diff
@@ -21,6 +21,31 @@ using ::testing::IsEmpty;
...
+// An `Event` for testing.
+// Example:
+// ```c++
+//    const int64_t create_time_ns = create_end - create_start;
+//    Int64Attr invocations("invocations", 1234);
+//    Int64Attr shader_hash("hash", res.shader_hash);
+//    std::vector<Attribute *> attributes = {&shader_hash, &invocations};
+//    TestEvent("create_shader_module_ns", attributes,
+//    LogLevel::High);
+// ```
```
kuhar: I would say that test classes defined and used in the same file do not require code samples: we can simply see how they are used in those tests. The compiler forces us to update those as the code changes, but there's nothing that tells us that comments are out-of-date.

kuhar (approved):

PR #137, opened 2022-11-25
Fix `gettid()` bug.

miladHakimi:
`glibc` versions earlier than 2.30 did not provide a wrapper around `gettid` system call. The workaround is using `syscall`.

### DIFF
--- a/layer/support/layer_utils.cc
+++ b/layer/support/layer_utils.cc
@@ -18,7 +18,7 @@
 #include <cstdint>
 
 #ifdef __linux__
-#include <sys/types.h>
+#include <sys/syscall.h>
 #include <unistd.h>
 #endif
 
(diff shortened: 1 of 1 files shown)


layer/support/layer_utils.cc
```diff
@@ -18,15 +18,15 @@
 #include <cstdint>
 
 #ifdef __linux__
-#include <sys/types.h>
+#include <sys/syscall.h>
```
vettoreldaniele, line 21: You're replacing `<sys/types.h>`. Is this intended?

miladHakimi: Yes.

vettoreldaniele: Why? `<sys/types.h>` is not where `gettid` was coming from.

kuhar: @vettoreldaniele `types.h` is not actually used anywhere in this file

vettoreldaniele: Ok so just a cleanup, that answers my question

vettoreldaniele (approved):

layer/support/layer_utils.cc
kuhar: So what conceptually happened is that Milad added a new header and separately dropped an unnecessary one.

PR #140, opened 2022-11-26
Fix formatting problem in float to string conversion.

miladHakimi:
`ostringstream` mixes the formatting of the floats. To represent floats, instead of relying on the stream to handle the float to string conversion, we use `std::to_chars` with `fixed` format.

The size of the string representing the number is unknown and varies from one input to the other. Hence, the input buffer given to `std::to_chars` is dynamically allocated and resized to fit  the number.

The size of the buffer might become more than the size of the characters written to it. Although all the rest of the characters are `'\0'`, writing it to a `ostringstream` copies the whole buffer because `ostringstream` uses the size of the string to perform the copy. By calling the `.data()`, the function copies the characters up until the first `'\0'` and does not copy the rest.

### DIFF
--- a/layer/support/trace_event_logging.cc
+++ b/layer/support/trace_event_logging.cc
@@ -12,16 +12,45 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#include "layer/support/trace_event_logging.h"
+
 #include <cassert>
+#include <charconv>
+#include <cstddef>
+#include <cstdint>
 #include <iomanip>
 #include <sstream>
 #include <string>
+#include <string_view>
 #include <vector>
 
 #include "layer/support/event_logging.h"
 #include "layer/support/layer_utils.h"
 
 namespace performancelayers {
+namespace {
+constexpr size_t kMaxBufferSize = 128;
+
+std::string DoubleToString(double value) {
+  std::string str;
+  // Start with the small-string size.
+  str.resize(str.capacity());
+  // Grow until we have enough space to store the number.
+  while (str.size() < kMaxBufferSize) {
+    auto [end, err] = std::to_chars(str.data(), str.data() + str.size(), value,
+                                    std::chars_format::fixed);
+    if (err == std::errc()) {
+      return str.data();
+    }
+    // Not enough space.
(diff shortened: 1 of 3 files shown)


layer/support/trace_event_logging.cc
```diff
@@ -38,12 +44,22 @@ std::string ValueToJsonString(const std::vector<int64_t> &values) {
   return json_str.str();
 }
 
-std::string ValueToJsonString(Duration value) {
-  return std::to_string(value.ToMilliseconds());
+std::string_view ValueToJsonString(std::array<char, kBufferSize> &buffer,
```
kuhar: Why not create a local `std::string`, use it as the buffer, and return at the end? You can get the pointer to the underlying storage by calling `.data()`

layer/support/trace_event_logging.h
```diff
@@ -15,13 +15,16 @@
...
 
+#include <array>
 #include <string>
 #include <vector>
 
 #include "layer/support/event_logging.h"
 #include "layer/support/log_output.h"
 
 namespace performancelayers {
+constexpr size_t kBufferSize = 32;
```
kuhar: How did you pick this constant?

kuhar (changes requested):

layer/support/trace_event_logging.cc
miladHakimi: Sure. Since the size of the underlying buffer can vary, using a dynamic allocation approach seems more reasonable.

layer/support/trace_event_logging.h
```diff
@@ -15,6 +15,7 @@
 #ifndef STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_TRACE_EVENT_LOGGING_H_
 #define STADIA_OPEN_SOURCE_PERFORMANCE_LAYERS_TRACE_EVENT_LOGGING_H_
 
+#include <array>
```
kuhar: nit: I don't think we need this anymore

kuhar (commented):
Just one nit

kuhar (approved):
LGTM, thanks!

PR #141, opened 2022-11-28
Add `TraceEventLogger` to the cache sideload layer.

miladHakimi:


layer/cache_sideload/cache_sideload_layer.cc
```diff
@@ -46,22 +51,26 @@ class CreateCacheEvent : public Event {
...
+        trace_attr_("trace_attr", "cache_sideload_layer", "i",
+                    {&scope_, &path_, &data_size_, &cache_size_}) {
+    InitAttributes({&path_, &data_size_, &cache_size_, &trace_attr_});
   }
 
  private:
   StringAttr path_;
   Int64Attr data_size_;
   Int64Attr cache_size_;
+  StringAttr scope_ = StringAttr("scope", "t");
```
kuhar: You can initialize it directly: `Type x(a, b)`

kuhar (commented):
Just one nit

layer/cache_sideload/cache_sideload_layer.cc
miladHakimi: I think I need to use curly brackets instead of parentheses.
https://stackoverflow.com/questions/24836526/why-cant-member-initializers-use-parentheses

kuhar (approved):

PR #143, opened 2022-11-28
Replace`DurationCLock::duration` with `Duration` class.

miladHakimi:
Although `Duration` is a wrapper around `DurationCLock::duration`, both co-existed in the code. This PR adds overloads to the `Duration` so that it can replace `DurationCLock::duration`.


kuhar (approved):
Nice cleanup!

PR #152, opened 2022-12-09
Ignore trace events in CSV logger

vettoreldaniele:
Currently, trace events are ignored by the CSV logger, but the delimiter appending logic does not take that into account, resulting in a delimiter being appended to every line with an empty value after. That causes CSV files not to match the declared header format anymore.

The invalid output looks like this:
```
Frame Time (ns),Benchmark State
16614739,1,
```

This PR makes it so that the CSV logger filters out events that it can't log (only trace events for now). Both unit and FileCheck tests are updated to check for this scenario.

The CommonLogger still outputs such events, but given the fact that it's somewhat more structured (outputs attributes in the format `attribute_name:attribute_value`), I feel it's ok if it outputs an attribute with an empty value as a parser has enough information to skip over it.

### DIFF
--- a/layer/support/csv_logging.cc
+++ b/layer/support/csv_logging.cc
@@ -17,7 +17,6 @@
 #include <sstream>
 #include <string>
 
-#include "layer/support/debug_logging.h"
 #include "layer/support/event_logging.h"
 #include "layer/support/layer_utils.h"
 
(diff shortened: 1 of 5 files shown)


test/check_compile_time_log.txt
```diff
@@ -1,4 +1,4 @@
 ; Checks the pattern of compile time log file. Makes sure the header
 ; and the data rows' format are as expected.
 ; CHECK-LABEL: Pipeline,Compile Time (ns)
-; CHECK-NEXT: "[{{0x[a-zA-Z0-9]+,0x[a-zA-Z0-9]+}}]",{{[0-9]+}}
+; CHECK-NEXT: "[{{0x[a-zA-Z0-9]+,0x[a-zA-Z0-9]+}}]",{{[0-9]+}}{{$}}
```
kuhar: Instead, we can run FileCheck with `--match-full-lines`

layer/support/csv_logging.cc
```diff
@@ -17,11 +17,15 @@
 #include <sstream>
 #include <string>
 
-#include "layer/support/debug_logging.h"
 #include "layer/support/event_logging.h"
 #include "layer/support/layer_utils.h"
 
 namespace performancelayers {
+namespace {
+const std::unordered_set<ValueType> kIgnoredEventTypes = {
```
kuhar: Since it's exactly one for now, can we just inline it? I do not foresee more types being ignored in near future.
I think a simple `if (... == ValueType::kTraceEvent) continue;` would do just fine.

layer/support/csv_logging.cc
```diff
@@ -57,44 +61,48 @@ std::string ValueToCSVString(Timestamp value) {
 std::string EventToCSVString(Event &event) {
   const std::vector<Attribute *> &attributes = event.GetAttributes();
 
+  // Remove events that should be ignored.
```
kuhar: s/events/attributes
For the context, this particular attribute name is a bit misleading -- it's a special attribute for TraceEvents, hence `TraceEventAttr`

layer/support/csv_logging.cc
vettoreldaniele: I removed the set and just checked the type directly in the filtering step, but I kept the filtering (instead of adding logic to look ahead when adding a delimiter and not adding it if it's `kTraceEvent`). This is basically to cleanly handle cases where the trace event attribute is in the middle or at the start of the attribute list or there are multiple trace events in the list. 

If you think that's never likely to happen, I can just rewrite this assuming that if there's a trace event, it's a single one at the end of the attribute list.

vettoreldaniele:
V1: Implemented suggestions. Changed FileCheck behavior to match full lines.

kuhar (approved):

PR #153, opened 2022-12-09
Add build instructions for the case where the Vulkan SDK is installed

vettoreldaniele:
This is to help users in the case where they are not building Vulkan Headers and Loaders from source but would rather use their installed SDK.

A few other header and typo fixes too.


kuhar (approved):
Thanks!

PR #154, opened 2022-12-17
Add filecheck for the trace event log file.

### LINKED ISSUE #149: [CI] Add `FileCheck`-based tests for the Chrome Trace Event format
Our current CI checks the format of the `CSV` and `CommonLog` files after each run to make sure the results are as expected. We need to have the same check for the Trace Event format as well.

miladHakimi:
This PR allows the CI to check the output of the trace event log file after the run is finished.

Fix: #149

### DIFF
--- a/docker/test_performance_layers.sh
+++ b/docker/test_performance_layers.sh
@@ -77,3 +77,6 @@ FileCheck "${PROJECT_ROOT_DIR}/test/check_frame_time_log.txt" \
 
 FileCheck "${PROJECT_ROOT_DIR}/test/check_event_log.txt" \
   --match-full-lines --input-file "${OUTPUT_DIR}"/events.log
+
+FileCheck "${PROJECT_ROOT_DIR}/test/check_trace_event.txt" \
+  --match-full-lines --input-file "${OUTPUT_DIR}"/trace_events.log
--- a/test/check_trace_event.txt
+++ b/test/check_trace_event.txt
@@ -1,0 +1,19 @@
+; Checks the pattern of the trace event log file. Makes sure the data rows
+; corresponding to each layer have the correct format.
+; Checks the shader hashes across different rows and guarantees they are
+; consistent.
+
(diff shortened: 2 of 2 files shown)


test/check_trace_event.txt
```diff
@@ -0,0 +1,18 @@
+; Checks the pattern of the trace event log file. Makes sure the data rows
+; Corresponding to each layer have the correct format.
```
kuhar: nit: C -> c

test/check_trace_event.txt
```diff
@@ -0,0 +1,18 @@
...
+; Checks the shader hashes across different rows and guarantees they are
+; consistent.
+; CHECK-DAG:  { "name" : "compile_time_layer_init", "ph" : "i", "cat" : "compile_time_layer", "pid" : {{[0-9]+}}, "tid" : {{[0-9]+}}, "ts" : {{[0-9]+\.[0-9]+}}, "s" : "g", "args" : { "scope" : "g" } },
```
kuhar, line 6: nit: typically we have an empty line before test description and checks

kuhar (commented):
Thank you @miladHakimi!

Does this run in CI? I'd think we have to add it to https://github.com/google/vulkan-performance-layers/blob/main/docker/test_performance_layers.sh#L66.

miladHakimi:
> Thank you @miladHakimi!
> 
> Does this run in CI? I'd think we have to add it to https://github.com/google/vulkan-performance-layers/blob/main/docker/test_performance_layers.sh#L66.

Sorry, I forgot to add it to the bash file. I think it should be fine now.

docker/test_performance_layers.sh
```diff
@@ -75,5 +75,8 @@ FileCheck "${PROJECT_ROOT_DIR}/test/check_memory_usage_log.txt" \
 FileCheck "${PROJECT_ROOT_DIR}/test/check_frame_time_log.txt" \
   --match-full-lines --input-file "${OUTPUT_DIR}"/frame_time.csv
 
-FileCheck "${PROJECT_ROOT_DIR}/test/check_event_log.txt" \
-  --match-full-lines --input-file "${OUTPUT_DIR}"/events.log
+FileCheck "${PROJECT_ROOT_DIR}/test/check_compile_time_log.txt" \
```
kuhar: This shouldn't be changed I think, we already have compile_time check at the top

miladHakimi: Sorry, my bad.

kuhar (approved):
Thanks!

PR #155, opened 2022-12-18
Fix the slack event bug.

### LINKED ISSUE #150: Fix Unaligned slack events
image
This is the output of the `vkcube` in the Trace Event format rendered by Perfetto UI. 

The `shader slack` events should start after the `shader creation` events (the small box in the bottom) and end before the `pipline creation` events. The start and end time is not measured properly in the layers.

miladHakimi:
A slack event shows the time between creation of a shader and its first use. `ShaderModuleSlack` is used to track these values using `creation_end_time` and `first_use_time` variables. Since the timestamp for an event shows the end of its creation, the `creation_end_time` should be set by the `CreateShaderEvent`'s timestamp. `first_use_time` should be set by the timestamp of the start of the pipeline.

Fix: #150


kuhar (approved):
Thank you!

PR #157, opened 2023-06-08
Added support for Layer Interface Version 0 in the pipeline runtime layer

mdagois:
Layer Interface Version 0 support + Android in the pipeline runtime layer

- Added support for Layer Interface Version 0, which is required for Android
- Fixed compilation errors under Android


layer/support/trace_event_logging.cc
```diff
@@ -33,6 +33,13 @@ constexpr size_t kMaxBufferSize = 128;
 
 std::string DoubleToString(double value) {
   std::string str;
+#if defined(__ANDROID__)
+  // stdlib on Android doesn't seem to support to_chars although the c++17
+  // flag should allow the compiler to use it (c++17 ops was removed).
+  std::ostringstream strs;
+  strs << value;
+  str = strs.str();
+#else
```
kuhar: IIRC we used to use stringstream but were run into issues with formats. Here, we'd want the format to be locale independent. Is there something else we could use instead like `sprintfn`?

I'd prefer if we used the same implementation for all targets. If that's not possible, could you move this alternative formatting to a new function and only guard the function call with ifdefs?

kuhar (commented):
Thanks for improving this, @mdagois!

This seems to only update the runtime layers -- did you omit the other layers intentionally? If so, could you update the PR description and title to mention that this is for the runtime layer only?

layer/runtime/runtime_layer.cc
```diff
@@ -23,13 +23,20 @@
...
 namespace {
 // ----------------------------------------------------------------------------
 // Layer book-keeping information
 // ----------------------------------------------------------------------------
 
+#define LAYER_NAME "VK_LAYER_STADIA_pipeline_runtime"
+
 constexpr uint32_t kRuntimeLayerVersion = 1;
-constexpr char kLayerName[] = "VK_LAYER_STADIA_pipeline_runtime";
+constexpr char kLayerName[] = LAYER_NAME;
```
kuhar: Do we need this define? I don't see it being used anywhere else

mdagois: Oops, sorry, I added the LAYER_NAME define and forgot to actually use it.

LAYER_NAME is now also used to build the entry point function names for Layer Interface Version 0 (see LAYER_NAME_FUNCTION macro).
I realized there were several issues around those macros, so I fixed them.
It requires a lot of macro gymnastics, but the intent is to have the layer name defined in one single place, because 
if we change the layer name later and forget to update the entry point names, there won't be any compilation error, but the layer won't work on Android anymore.

I hope that makes sense.

mdagois:
> This seems to only update the runtime layers -- did you omit the other layers intentionally? If so, could you update the PR description and title to mention that this is for the runtime layer only?

Yes, it was intentional.
My intent is to eventually add support to all layers once we get that one fully working.
I updated the description and title of the PR.

layer/support/trace_event_logging.cc
mdagois: Ah, I see. The comma probably gets in the way. Isn't `sprintf` afflicted by the same issue? it seems I need to set the locale, which is not thread-safe. I doubt many Vulkan app would care, but who knows.

Do you think it would be reasonable to make a custom function?
DoubleToString is used only in that file.
It is only dealing with doubles because we want milliseconds instead of nanoseconds.
I am thinking about a function that would convert nanoseconds (int64_t) to a string of milliseconds.
For example, 1234567890 would become "1234.567890".

Something roughly like this:

```
std::string IntNanosecondToDoubleMicrosecondString(int64_t value) {
  std::ostringstream ss;
  ss << value / 1000000 << "." << std::setw(6) << std::setfill('0') << value % 1000000;
  return ss.str();
}
```

I would need to change the tests.
Please let me know if I am overthinking the issue.

layer/runtime/runtime_layer.cc
```diff
@@ -483,3 +493,65 @@ SPL_LAYER_ENTRY_POINT SPL_RUNTIME_LAYER_FUNC(PFN_vkVoidFunction,
...
+
+#define LAYER_NAME_FUNCTION_CONCAT(layername, func) layername##func
+#define LAYER_NAME_FUNCTION_CONCAT_EXPAND(layername, func) \
+  LAYER_NAME_FUNCTION_CONCAT(layername, func)
+#define LAYER_NAME_FUNCTION(func) \
+  LAYER_NAME_FUNCTION_CONCAT_EXPAND(LAYER_NAME, func)
+
+// Exposes the layer interface version 0's GetInstanceProcAddr
+SPL_LAYER_ENTRY_POINT VKAPI_ATTR PFN_vkVoidFunction VKAPI_CALL
+LAYER_NAME_FUNCTION(GetInstanceProcAddr)(VkInstance instance,
```
kuhar: What do you think about one more macro to define an entry point function with a given name and type? I would see it used like this:

```c++
SPL_RUNTIME_LAYER_INTERFACE(PFN_vkvoidFunvtion, GetInstanceProcAddr)((VkInstance instance,
```

That would add all the other macros with function attributes

layer/support/trace_event_logging.cc
kuhar: The property I care about the most is being able to produce a trace log on one device and read it on another device, getting the same results on both ends. Not sure what the best way to handle floats is -- if a simple ostringstream unblocks you, this is fine with me as long as there's a comment explaining the reasoning behind it and/or a TODO for platform-agnostic float printing.

layer/runtime/runtime_layer.cc
kuhar: We have something similar for the existing functions above `SPL_RUNTIME_LAYER_FUNC`

layer/support/trace_event_logging.cc
mdagois: OK, so I switched to a very simple custom implementation that is platform-agnostic and locale-independent. It should be in the format that most humans would use. It generates redundant zeros (e.g. 123.456000) in a fully predictable manner, so I updated the tests cases to reflect that.
I also wrote a note that to_chars should be used once Android supports it.

layer/runtime/runtime_layer.cc
mdagois: I added a new macro `SPL_RUNTIME_LAYER_INTERFACE` that uses the same arguments as `SPL_RUNTIME_LAYER_FUNC` does (i.e. return type, func name and args).
I use the macro for the two entry points that require the layer name prepended.

layer/support/trace_event_logging.cc
kuhar: Perfect, thank you!

kuhar (approved):
LGTM

